PR #3, opened 2020-12-19
Resolve check annotations

jonlouie:
*Issue #, if available:*

*Description of changes:*
* Resolved all compiler warnings to 1) clean up dev environment and 2) resolve check annotations in Github
* Delete unused namespace references

### DIFF
--- a/src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
+++ b/src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
@@ -43,22 +43,22 @@ namespace CTA.Rules.Actions
         /// <returns></returns>
         private configuration ProcessWebConfig(string projectDir)
         {
-            string webConfigFileDir = Path.Combine(projectDir, "web.config");
+            string webConfigFile = Path.Combine(projectDir, "web.config");
 
             configuration webConfig = null;
-            if (File.Exists(webConfigFileDir))
+            if (File.Exists(webConfigFile))
             {
                 try
                 {
-                    using (FileStream reader = File.OpenRead(webConfigFileDir))
+                    using (FileStream reader = File.OpenRead(webConfigFile))
                     {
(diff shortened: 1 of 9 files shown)


src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
```diff
@@ -56,7 +56,7 @@ private configuration ProcessWebConfig(string projectDir)
                         webConfig = (configuration)ser.Deserialize(reader);
                     }
                 }
-                catch (Exception ex)
```
marknfawaz, line 59: LogError accepts an exception parameter. can we change to that instead. something like:
LogHelper.LogError(ex, string.Format("Error processing web.config file {0}", webConfigFileDir))

src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -178,7 +178,7 @@ private void CheckResources()
                     File.WriteAllBytes(zipFile, fileContents);
                     ZipFile.ExtractToDirectory(zipFile, executingPath, true);
                 }
-                catch (Exception ex)
+                catch (Exception)
                 {
```
marknfawaz, line 182: Same as above. can we log the error?

src/CTA.Rules.RuleFiles/RulesFileParser.cs
```diff
@@ -587,7 +584,7 @@ private void ParseActions(NodeToken nodeToken, List<Action> actions)
                             }
                     }
                 }
-                catch (Exception ex)
+                catch (Exception)
                 {
```
marknfawaz, line 588: Same as above. Let's log the error. If we have to, change to LogError

marknfawaz (changes requested):
Looks good. minor changes..

marknfawaz (approved):

PR #9, opened 2020-12-31
add aspnet core project type features

jonlouie:
*Issue #, if available:*

*Description of changes:*
* add project type feature for ASP.NET Core MVC projects
* add project type feature for ASP.NET Core WebApi projects
* Notes RE: how these features are detected
  * MVC conditions: 
     1. At least 1 controller derived from `Controller`<sup>1</sup> abstract class that also returns a View object<sup>2</sup>
     2. AND `Views` folder exists and is not empty
  * WebApi conditions: 
    1. At least 1 controller derived from `ControllerBase`<sup>1</sup> with zero controllers derived from `Controller` abstract class
    2. OR all controllers derived from `Controller` abstract class do not return any View objects

<sup>1</sup> ASP.NET Core merged MVC and WebApi libraries in the same MVC library so we need to be more mindful of how the libraries are use rather than looking for package references like we did for ASP.NET project types. The `ControllerBase` abstract class is generally used for WebApi projects as they do not have any methods that return View objects. The `Controller` abstract class is generally used for MVC controllers and allows for returning View objects, but they don't always have to return views. Core WebApi projects that do not follow best practices might use a controller derived from the `Controller` abstract class so we want to account for those cases.

<sup>2</sup> "View objects" in this context are any instances from a class that implements `IActionResult` and also returns a View or a Partial View (i.e. `PartialViewResult`, `ViewResult`, or `ViewComponentResult`). These classes were determined from the [list of classes derived from the IActionResult interface](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.iactionresult?view=aspnetcore-3.1)

---

### DIFF
--- a/src/CTA.FeatureDetection.ProjectType/CompiledFeatures/AspNetCoreWebApiFeature.cs
+++ b/src/CTA.FeatureDetection.ProjectType/CompiledFeatures/AspNetCoreWebApiFeature.cs
@@ -1,0 +1,33 @@
+﻿using Codelyzer.Analysis;
+using CTA.FeatureDetection.Common.Extensions;
+using CTA.FeatureDetection.Common.Models.Features.Base;
+using System.Linq;
+
+namespace CTA.FeatureDetection.ProjectType.CompiledFeatures
+{
+    public class AspNetCoreWebApiFeature : CompiledFeature
+    {
+        /// <summary>
+        /// Determines that a project is an ASP.NET Core WebAPi project if:
+        ///     1) There are any classes derived from the ControllerBase abstract class
+        ///        and no classes derived from the Controller abstract class
+        ///     OR
+        ///     2) If there are any classes derived from the Controller abstract class,
+        ///        none of them call any methods that return a view-related object.
+        /// </summary>
+        /// <param name="analyzerResult"></param>
+        /// <returns>Whether a project is an ASP.NET Core WebApi project or not</returns>
+        public override bool IsPresent(AnalyzerResult analyzerResult)
+        {
+            var project = analyzerResult.ProjectResult;
+            var controllerBaseClassDeclarations = project.GetClassDeclarationsWithBaseType(Constants.NetCoreMvcControllerBaseOriginalDefinition);
+            var controllerClassDeclarations = project.GetClassDeclarationsWithBaseType(Constants.NetCoreMvcControllerOriginalDefinition);
+            var returnsViewObjectFromControllerClasses = controllerClassDeclarations.Any(c => c.GetInvocationExpressionsBySemanticReturnType(Constants.NetCoreViewResultTypes).Any());
+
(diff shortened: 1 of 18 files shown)


src/CTA.FeatureDetection.ProjectType/CompiledFeatures/AspNetCoreWebApiFeature.cs
```diff
@@ -0,0 +1,33 @@
...
+        ///        none of them call any methods that return a view-related object.
+        /// </summary>
+        /// <param name="analyzerResult"></param>
+        /// <returns>Whether a project is an ASP.NET Core WebApi project or not</returns>
+        public override bool IsPresent(AnalyzerResult analyzerResult)
+        {
+            var project = analyzerResult.ProjectResult;
+            var controllerBaseClassDeclarations = project.GetClassDeclarationsWithBaseType(Constants.NetCoreMvcControllerBaseOriginalDefinition);
+            var controllerClassDeclarations = project.GetClassDeclarationsWithBaseType(Constants.NetCoreMvcControllerOriginalDefinition);
+            var returnsViewObjectFromControllerClasses = controllerClassDeclarations.Any(c => c.GetInvocationExpressionsBySemanticReturnType(Constants.NetCoreViewResultTypes).Any());
```
marknfawaz, line 25: How does this help us detect a web api? When a controller is returning a View, usually it means it's an Mvc project.

One thing we can look for is the ApiController attribute in a class declaration, although im not sure if that covers all cases. I will do some research too.

marknfawaz (changes requested):
Looks ok to me, just need to further discuss the web api detection mechanism.

src/CTA.FeatureDetection.ProjectType/CompiledFeatures/AspNetCoreWebApiFeature.cs
jonlouie: We are looking for Views being returned because it's possible that a WebApi project will use an MVC controller. This part of the condition is: If there are any MVC Controllers, make sure they are NOT returning any Views (partial or otherwise).

I would definitely appreciate additional research here as well. I think the line between WebApi and MVC projects is much more grey in .NET Core since Microsoft merged the two libraries.

marknfawaz: Here's what I found:

- All projects, regardless of type (Mvc or web API) will have a non-empty views folder initially, so we can't use views as a condition
- An Mvc project can also be a web api project, and vice versa
- For MVC: Return should be a view. Not sure if you checked, but the return type of the below should be View, not IActionResult
`
public IActionResult Index()
{
    return View();
}
`
- For Web API, the condition should be: 1) Controller with ApiController attribute (regardless of what it inherits from) or 2) Controller with a public method that doesn't return a view.

marknfawaz (approved):

PR #10, opened 2021-01-05
Add new metrics and update sleet version

marknfawaz:
Add metrics for project target version, package updates and package actions.
Update sleet version


jonlouie (approved):
LGTM

src/CTA.Rules.Metrics/Models/BuildErrorMetric.cs
```diff
@@ -0,0 +1,50 @@
...
+            ProjectGuid = context.ProjectGuidMap.GetValueOrDefault(projectPath, "N/A");
+        }
+
+        private string ExtractBuildErrorCode(string buildError)
+        {
+            var pattern = new Regex(BuildErrorCodePattern);
+            var matches = pattern.Matches(buildError);
+            var errorCode = matches.FirstOrDefault(m => buildError.StartsWith(m.Value))?.Value;
+
+            return errorCode ?? string.Empty;
```
birojnayak, line 47: if it doesn't start with "cs", we can still keep it ? what's wrong with that..

src/CTA.Rules.Metrics/Models/CTAMetric.cs
```diff
@@ -0,0 +1,10 @@
+﻿using Newtonsoft.Json;
+
+namespace CTA.Rules.Metrics
+{
+    public abstract class CTAMetric
```
birojnayak, line 5: why abstract ? you don't have any abstract method..

src/CTA.Rules.Metrics/Models/DownloadedFilesMetric.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    public class DownloadedFilesMetric : CTAMetric
+    {
+        [JsonProperty("metricName", Order = 10)]
+        public string MetricName => "RulesFile";
+
+        [JsonProperty("downloadedFile", Order = 11)]
+        public string DownloadedFile { get; set; }
+
+        [JsonProperty("solutionPath", Order = 12)]
```
birojnayak, line 13: do you think "solutionPath" / "projectguid" etc (used more than once), could be moved to a constant file ?

src/CTA.Rules.Metrics/MetricsTransformer.cs
```diff
@@ -26,6 +26,46 @@ internal static IEnumerable<GenericActionExecutionMetric> TransformGenericAction
...
+        }
+
+        internal static IEnumerable<UpgradePackageMetric> TransformUpgradePackages(MetricsContext context, ProjectResult projectResult)
+        {
+            var projectFile = projectResult.ProjectFile;
+            var upgradePackages = projectResult.UpgradePackages;
+            var upgradePackagesMetric = new List<UpgradePackageMetric>();
+            foreach (var upgradePackage in upgradePackages)
+            {
+                upgradePackagesMetric.Add(new UpgradePackageMetric(context, upgradePackage, projectFile));
```
birojnayak, line 49: is there a way we can also get source package version ? and in the metrics just add "a.b.c(v1.0)" replaced by "a.b.e(v.3.4)" something in this line ?

src/CTA.Rules.Metrics/PortSolutionResultReportGenerator.cs
```diff
@@ -45,20 +48,32 @@ private void GenerateMetrics()
...
             var actionExecutionMetrics = new List<GenericActionExecutionMetric>();
+
             foreach (var projectResult in PortSolutionResult.ProjectResults)
             {
-                var executedActionMetrics = MetricsTransformer.TransformGenericActionExecutions(Context, projectResult);
-                actionExecutionMetrics.AddRange(executedActionMetrics);
+                targetVersionMetrics.AddRange(MetricsTransformer.TransformTargetVersions(Context, projectResult));
+                upgradePackageMetrics.AddRange(MetricsTransformer.TransformUpgradePackages(Context, projectResult));
+                actionPackageMetrics.AddRange(MetricsTransformer.TransformActionPackages(Context, projectResult));
+                actionExecutionMetrics.AddRange(MetricsTransformer.TransformGenericActionExecutions(Context, projectResult));
```
birojnayak, line 61: if these methods are time consuming you can make them async.. if not time consuming ok to keep in this way...

src/CTA.Rules.Metrics/PortSolutionResultReportGenerator.cs
```diff
@@ -45,20 +48,32 @@ private void GenerateMetrics()
...
             GenericActionExecutionMetrics = actionExecutionMetrics;
-        }
+        } 
 
         private string GeneratePortSolutionResultJsonReport()
         {
             var allMetrics = new List<CTAMetric>();
             allMetrics = allMetrics.Concat(ReferencesMetrics)
                 .Concat(DownloadedFilesMetrics)
+                .Concat(TargetVersionMetrics)
```
birojnayak, line 74: ensure with additional metrics going as json, there is no performance bottleneck... meaning when we are uploading these json, all those should be async way... else see if you can point me the code where it's being uploaded to s3..

src/CTA.Rules.Metrics/Models/BuildErrorMetric.cs
jonlouie: For `BuildErrorMetric.BuildErrorCode`, we are parsing an error message string to get the code.  Based on [the docs for msbuild](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/), we try parsing out the patten `CS####`. 

If we get a build error message that does not contain an error code with this pattern, we still log the error, it just won't have a code associated with it.

src/CTA.Rules.Metrics/Models/CTAMetric.cs
jonlouie: This is abstract to prevent developers from instantiating this class directly. This serves as a starting template for all of our metrics classes, but this class itself should never be written instantiated.

[Here is an example of how we use this common base class to our advantage](https://github.com/aws/cta/blob/main/src/CTA.Rules.Metrics/PortSolutionResultReportGenerator.cs#L59). We aggregate all of our metrics objects into a single collection (using this class as the generic type) to generate a report.

src/CTA.Rules.Metrics/Models/DownloadedFilesMetric.cs
jonlouie: Generally, I agree that string literals should be stored in a constants file, but these strings are used to rename these properties when the instances of the class are serialized to json (we do this when we are exporting the metrics to logs/files). Since the output json is decoupled from the application itself, I don't think there would be much reusability in storing these string values as constants.

src/CTA.Rules.Metrics/PortSolutionResultReportGenerator.cs
jonlouie: These metrics are stored in log messages which are processed by the Porting Assistant logger, so the way these metrics are written to disk and stored to S3 are the same as how Porting Assistant uploads logs to S3... Unfortunately, I'm not sure where in the PA code that is done 🙂

jonlouie:
@birojnayak FYI: some of your comments are related to decisions I made in the first iteration of our CTA metrics so I followed up on those.

src/CTA.Rules.Metrics/MetricsTransformer.cs
marknfawaz: Yes. already started to work on this with PA. I will create a new PR to address the new metrics.

src/CTA.Rules.Metrics/Models/CTAMetric.cs
birojnayak: so you make them internal..

src/CTA.Rules.Metrics/Models/DownloadedFilesMetric.cs
birojnayak: the point is when you are changing "solutionPath" to "SolutionFullPath" across all metrics.. you might have to change across many CS files instead of one.... so evaluate and make changes if needed

src/CTA.Rules.Metrics/MetricsTransformer.cs
birojnayak: ok create new Issue and track and resolve the comment

src/CTA.Rules.Metrics/Models/DownloadedFilesMetric.cs
jonlouie: Oh, I understand now. Making the property name uniform across multiple metrics would warrant a change here. Thank you for pointing that out. I can open an issue and create a separate PR for that.

src/CTA.Rules.Metrics/Models/CTAMetric.cs
jonlouie: Making the class internal would make sense here (it would also make sense to change all metrics object classes to internal), but within this assembly we still would want to avoid instantiating a `CTAMetric` metric directly.

PR #22, opened 2021-01-25
Owin method modifier action

Eruanion:
*Issue #, if available:*
We need a new rule to replace method modifiers.

*Description of changes:*

Added the new class declaration action for replacing method modifiers for passed in method name and modifier list. Included unit tests as well.

### DIFF
--- a/src/CTA.Rules.Actions/ClassActions.cs
+++ b/src/CTA.Rules.Actions/ClassActions.cs
@@ -1,10 +1,12 @@
 ﻿using CTA.Rules.Config;
+using CTA.Rules.Models;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Classification;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Editing;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace CTA.Rules.Actions
@@ -138,10 +140,10 @@ namespace CTA.Rules.Actions
                 var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
                 if(allMethods.Count() > 0)
                 {
-                    var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName);
-                    if(removeMethod.Count() > 0)
+                    var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
+                    if(removeMethod != null)
                     {
-                        node = node.RemoveNode(removeMethod.FirstOrDefault(), SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
+                        node = node.RemoveNode(removeMethod, SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
                     }
                 }
                 
@@ -166,13 +168,19 @@ namespace CTA.Rules.Actions
                 var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
                 if(allMethods.Count() > 0)
                 {
-                    var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName);
-                    if(replaceMethod.Count() > 0)
+                    var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
+                    if(replaceMethod != null )
                     {
-                        SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(modifiers));
-                        var newMethod = replaceMethod.FirstOrDefault().WithModifiers(tokenList);
+                        string[] modifierList = modifiers.Split(new char[] { ' ', ','});
+                        var verifiedModifiers = modifierList.Where(m => CTA.Rules.Config.Constants.supportedMethodModifiers.Contains(m.ToLower()));
+                        if(modifierList.Count() == verifiedModifiers.Count())
+                        {
+                            string supportedModifier = String.Join(" ", verifiedModifiers);
+                            SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(supportedModifier));
+                            var newMethod = replaceMethod.WithModifiers(tokenList);
 
-                        node = node.WithMembers(node.Members.Replace(replaceMethod.FirstOrDefault(), newMethod)).NormalizeWhitespace();
+                            node = node.WithMembers(node.Members.Replace(replaceMethod, newMethod)).NormalizeWhitespace();
+                        }
                     }
                 }
 
--- a/src/CTA.Rules.Actions/ClassActions.cs
+++ b/src/CTA.Rules.Actions/ClassActions.cs
@@ -136,9 +136,15 @@ namespace CTA.Rules.Actions
             {
                 var allMembers = node.Members.ToList();
                 var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
-                var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
-                allMembers.Remove(removeMethod);
-                node = node.RemoveNode(removeMethod, SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
+                if(allMethods.Count() > 0)
+                {
+                    var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName);
+                    if(removeMethod.Count() > 0)
+                    {
+                        node = node.RemoveNode(removeMethod.FirstOrDefault(), SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
+                    }
+                }
+                
                 return node;
             };
             return AddMethod;
@@ -158,13 +164,17 @@ namespace CTA.Rules.Actions
             {
                 var allMembers = node.Members.ToList();
                 var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
-                var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
-                allMembers.Remove(replaceMethod);
-                SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(modifiers));
-                var newMethod = replaceMethod.WithModifiers(tokenList);
-                allMembers.Add(replaceMethod);
+                if(allMethods.Count() > 0)
+                {
+                    var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName);
+                    if(replaceMethod.Count() > 0)
+                    {
+                        SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(modifiers));
+                        var newMethod = replaceMethod.FirstOrDefault().WithModifiers(tokenList);
 
-                node = node.WithMembers(node.Members.Replace(replaceMethod, newMethod)).NormalizeWhitespace();
+                        node = node.WithMembers(node.Members.Replace(replaceMethod.FirstOrDefault(), newMethod)).NormalizeWhitespace();
+                    }
+                }
 
                 return node;
             };
--- a/src/CTA.Rules.Actions/ClassActions.cs
+++ b/src/CTA.Rules.Actions/ClassActions.cs
@@ -138,9 +138,9 @@ namespace CTA.Rules.Actions
             {
                 var allMembers = node.Members.ToList();
                 var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
-                if(allMethods.Count() > 0)
+                if(allMethods.Any())
                 {
-                    var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
+                    var removeMethod = allMethods.FirstOrDefault(m => m.Identifier.ToString() == methodName);
                     if(removeMethod != null)
                     {
                         node = node.RemoveNode(removeMethod, SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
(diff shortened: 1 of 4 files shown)


src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +152,23 @@ public class ClassActions
...
+            {
+                var allMembers = node.Members.ToList();
+                var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
+                var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
+                allMembers.Remove(replaceMethod);
+                SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(modifiers));
+                var newMethod = replaceMethod.WithModifiers(tokenList);
+                allMembers.Add(replaceMethod);
+
+                node = node.WithMembers(node.Members.Replace(replaceMethod, newMethod)).NormalizeWhitespace();
```
marknfawaz, line 167: Why are we modifying allMembers if we're just replacing it. We can either use allMembers inside the WithMembers, or not modify it.

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +152,23 @@ public class ClassActions
...
             return RenameClass;
         }
+        public Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> GetReplaceMethodModifiersAction(string methodName, string modifiers)
+        {
+            Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> ReplaceMethodModifiers = (SyntaxGenerator syntaxGenerator, ClassDeclarationSyntax node) =>
+            {
+                var allMembers = node.Members.ToList();
+                var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
+                var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
+                allMembers.Remove(replaceMethod);
```
marknfawaz, line 162: This will probably fail if it runs on an empty class

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +152,23 @@ public class ClassActions
             };
             return RenameClass;
         }
+        public Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> GetReplaceMethodModifiersAction(string methodName, string modifiers)
+        {
+            Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> ReplaceMethodModifiers = (SyntaxGenerator syntaxGenerator, ClassDeclarationSyntax node) =>
+            {
+                var allMembers = node.Members.ToList();
+                var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
+                var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
```
marknfawaz, line 161: Can m.Identifier be null or empty?

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +152,23 @@ public class ClassActions
...
+                allMembers.Remove(replaceMethod);
+                SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(modifiers));
+                var newMethod = replaceMethod.WithModifiers(tokenList);
```
marknfawaz, line 164: What if the class or empty, or we didn't find the method in the class? Can replaceMethod be null?

marknfawaz (changes requested):

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Copied from the remove action, I went ahead and fixed that one too :-) thanks.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Sorry, good catch I fixed it and also fixed RemoveMethod action which this was based on.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Not anymore after the above fixes.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: It could, this has been fixed as well after you mentioned the first 2 problems. Thanks.

Eruanion (commented):
Thanks Mark, good catches.

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -136,9 +136,15 @@ public class ClassActions
...
-                var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
-                allMembers.Remove(removeMethod);
-                node = node.RemoveNode(removeMethod, SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
+                if(allMethods.Count() > 0)
+                {
+                    var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName);
+                    if(removeMethod.Count() > 0)
+                    {
+                        node = node.RemoveNode(removeMethod.FirstOrDefault(), SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
+                    }
```
marknfawaz, line 145: nit: When getting removeMethod, use FirstOrDefault, then check if it's null. This way you won't have to call FirstOrDefault again inside the condition.

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +158,27 @@ public class ClassActions
...
+                if(allMethods.Count() > 0)
+                {
+                    var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName);
+                    if(replaceMethod.Count() > 0)
+                    {
+                        SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(modifiers));
+                        var newMethod = replaceMethod.FirstOrDefault().WithModifiers(tokenList);
+
+                        node = node.WithMembers(node.Members.Replace(replaceMethod.FirstOrDefault(), newMethod)).NormalizeWhitespace();
+                    }
```
marknfawaz, line 176: nit: When getting replaceMethod, use FirstOrDefault, then check if it's null. This way you won't have to call FirstOrDefault twice inside the condition.

marknfawaz (approved):
A couple minor comments. Feel free to change or merge as is.

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +158,27 @@ public class ClassActions
             };
             return RenameClass;
         }
+        public Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> GetReplaceMethodModifiersAction(string methodName, string modifiers)
+        {
+            Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> ReplaceMethodModifiers = (SyntaxGenerator syntaxGenerator, ClassDeclarationSyntax node) =>
```
birojnayak, line 163: just ensure modifiers are in a pre-defined List {void,public,private etc} ..

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Re-tested as mentioned and updated.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Re-tested as mentioned and updated.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Agreed, added to the constant file and implemented a check to verify modifiers being sent in are all supported.

birojnayak (dismissed):
looks ok

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -136,9 +138,15 @@ public class ClassActions
             {
                 var allMembers = node.Members.ToList();
                 var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
-                var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
-                allMembers.Remove(removeMethod);
-                node = node.RemoveNode(removeMethod, SyntaxRemoveOptions.KeepNoTrivia).NormalizeWhitespace();
+                if(allMethods.Count() > 0)
```
jonlouie, line 141: nit: can replace condition with `allMethods.Any()`

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -136,9 +138,15 @@ public class ClassActions
...
+                {
+                    var removeMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
```
jonlouie, line 143: nit: can replace `.Where(someCondition).FirstOrDefault()` with `.FirstOrDefault(someCondition)`

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +160,33 @@ public class ClassActions
             };
             return RenameClass;
         }
+        public Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> GetReplaceMethodModifiersAction(string methodName, string modifiers)
+        {
+            Func<SyntaxGenerator, ClassDeclarationSyntax, ClassDeclarationSyntax> ReplaceMethodModifiers = (SyntaxGenerator syntaxGenerator, ClassDeclarationSyntax node) =>
+            {
+                var allMembers = node.Members.ToList();
+                var allMethods = allMembers.OfType<MethodDeclarationSyntax>();
+                if(allMethods.Count() > 0)
```
jonlouie, line 169: nit: same as above, can replace with `.Any()`

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +160,33 @@ public class ClassActions
...
+                {
+                    var replaceMethod = allMethods.Where(m => m.Identifier.ToString() == methodName).FirstOrDefault();
```
jonlouie, line 171: nit: same as above, can simplify to `.FirstOrDefault(condition)`

src/CTA.Rules.Config/Constants.cs
```diff
@@ -89,5 +89,7 @@ public class Constants
         public const string key = "key";
         public const string value = "value";
         public const string ConnectionStrings = "ConnectionStrings";
+
+        public static readonly List<string> supportedMethodModifiers = new List<string>() { "public", "internal", "protected", "private", "abstract", "extern", "override", "static", "unsafe", "virtual", "async" };
```
jonlouie, line 93: nit: capitalize static variables/constants

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +160,33 @@ public class ClassActions
...
+                    if(replaceMethod != null )
+                    {
+                        string[] modifierList = modifiers.Split(new char[] { ' ', ','});
+                        var verifiedModifiers = modifierList.Where(m => CTA.Rules.Config.Constants.supportedMethodModifiers.Contains(m.ToLower()));
+                        if(modifierList.Count() == verifiedModifiers.Count())
```
jonlouie, line 176: Is it safe to use `.ToLower()` here? Modifiers are case-sensitive so this might allow something like `INTERNAL MyMethod(...)`

Also a suggestion that is open for debate: Instead of comparing the counts to validate, we can replace `.Where(...)` with `All(...)`. It will return a boolean stating whether or not all items in the collection match the condition. Up to you if you think this will improve readability. To add to this, validation can be placed in a separate method if you believe it will improve readability and/or if the validation logic is likely to be reused.

src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +160,33 @@ public class ClassActions
...
+                        {
+                            string supportedModifier = String.Join(" ", verifiedModifiers);
+                            SyntaxTokenList tokenList = new SyntaxTokenList(SyntaxFactory.ParseTokens(supportedModifier));
+                            var newMethod = replaceMethod.WithModifiers(tokenList);
```
jonlouie, line 180: If above suggestion is taken, we can probably replace `supportedModifier` with `modifierList`.

jonlouie (commented):
Mostly styling nits, but there is one logic questions regarding the use of `.ToLower()`

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: nice, implemented as suggested.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Also implemented as suggested.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Implemented a suggested, thanks :-)

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: Also implemented as suggested.

src/CTA.Rules.Config/Constants.cs
Eruanion: Capitalized as suggested, we should make the other variables also match this requirement.

jonlouie (approved):
LGTM!

marknfawaz (approved):

PR #31, opened 2021-02-02
Add support for previous version when upgrading packages

### LINKED ISSUE #29: Issue when packages are upgraded using PA
When packages are selected in PA for upgrade, the csproj conversion process is failing and causing the project file to stay .net framework

marknfawaz:
*Issue #, if available:*
Closes #29 

*Description of changes:*
This change adds the ability to log the source and target versions when upgrading packages


src/CTA.Rules.Models/ProjectConfiguration.cs
```diff
@@ -10,13 +10,13 @@ public class ProjectConfiguration
         public ProjectConfiguration()
         {
             TargetVersions = new List<string> { Constants.DefaultCoreVersion };
-            PackageReferences = new Dictionary<string, string>();
+            PackageReferences = new Dictionary<string, Tuple<string, string>>();
```
jonlouie, line 13: Is there a reason for using `Tuple<string, string>` instead of an object? My thought is that since the original and new versions are always coupled, using an object to contain them could improve readability.

src/CTA.Rules.Models/Actions/Packageaction.cs
```diff
@@ -7,6 +7,7 @@ namespace CTA.Rules.Models
     public class PackageAction
     {
         public string Name { get; set; }
+        public string PreviousVersion { get; set; }
```
jonlouie, line 10: nit: consider renaming to `OriginalVersion` to match the property name in `UpgradePackageMetric` and also reduce ambiguity as to what the value represents (i.e. `PreviousVersion` could also mean "the version released before `PackageVersion`", but `OriginalVersion` has fewer interpretations IMHO).

jonlouie (commented):
Left 2 minor comments, otherwise looking pretty good 👍

src/CTA.Rules.Metrics/PortSolutionResultReportGenerator.cs
```diff
@@ -52,24 +76,34 @@ private void GenerateMetrics()
...
+            var allMetrics = new List<CTAMetric>();
+            allMetrics = allMetrics
+                .Concat(UpgradePackageMetrics)
+                .Concat(ActionPackageMetrics)
+                .Concat(GenericActionMetrics)
+                .ToList();
+
+            PortSolutionResultJsonReport = JsonConvert.SerializeObject(allMetrics);
+            return PortSolutionResultJsonReport;
+        }
```
jonlouie, line 106: Do you think there is any risk sharing the `PortSolutionResultJsonReport` property with the other report-generating method `GeneratePortSolutionResultJsonReport()`?

This also might be an indicator that this class should be refactored such that we have static methods to generate and export the different reports. If you believe that's the case, then we can save the refactor for another issue + PR.

tst/CTA.Rules.Test/Metrics/PortSolutionResultReportGeneratorTests.cs
```diff
@@ -246,30 +246,11 @@ public void GenerateAndExportReports_Creates_Expected_Json_Report()
...
-  {
-    ""metricsType"": ""CTA"",
-    ""metricName"": ""GenericAction"",
-    ""actionName"": ""GA2 Name"",
-    ""actionType"": ""GA2 Type"",
-    ""actionValue"": ""GA2 Value"",
-    ""solutionPath"": ""5fa9de0cb5af2d468dfb1702b1e342f47de2df9a195dabb3be2d04f9c2767482"",
-    ""projectGuid"": ""1234-5678"",
-    ""filePath"": ""eb98c1d648bc61064bdeaca9523a49e51bb3312f28f59376fb385e1569c77822""
-  },
```
jonlouie, line 272: Should we have a unit test for the analysis metrics? It could help document the format of the metrics being logged.

jonlouie (dismissed):
LGTM 👍

jonlouie (approved):

Issue #32, opened 2021-02-04
Add unit tests for Antlr porting support


jonlouie:
Write unit tests to support the rules mentioned in this issue: https://github.com/aws/porting-assistant-dotnet-datastore/issues/28

PR #33, opened 2021-02-04
Issue 32

jonlouie:
*Issue #, if available:* https://github.com/aws/cta/issues/32

*Description of changes:*
* Add unit tests for porting rules added here: 
* Add property initialization to prevent NPE's

### DIFF
--- a/src/CTA.Rules.Models/ProjectActions.cs
+++ b/src/CTA.Rules.Models/ProjectActions.cs
@@ -14,6 +14,7 @@ namespace CTA.Rules.Models
             FileActions = new BlockingCollection<FileActions>();
             PackageActions = new BlockingCollection<PackageAction>();
             ProjectReferenceActions = new BlockingCollection<string>();
+            ProjectLevelActions = new List<ProjectLevelAction>();
         }
         public BlockingCollection<FileActions> FileActions { get; set; }
         public BlockingCollection<PackageAction> PackageActions { get; set; }
--- a/tst/CTA.Rules.Test/PortCoreTests.cs
+++ b/tst/CTA.Rules.Test/PortCoreTests.cs
@@ -2,7 +2,6 @@ using CTA.Rules.PortCore;
 using NUnit.Framework;
 using System.IO;
 using System.Linq;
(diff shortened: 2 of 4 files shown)


src/CTA.Rules.Models/ProjectActions.cs
```diff
@@ -14,6 +14,7 @@ public ProjectActions()
             FileActions = new BlockingCollection<FileActions>();
             PackageActions = new BlockingCollection<PackageAction>();
             ProjectReferenceActions = new BlockingCollection<string>();
+            ProjectLevelActions = new List<ProjectLevelAction>();
```
marknfawaz, line 17: Do we need this?

tst/CTA.Rules.Test/PortCoreTests.cs
```diff
@@ -276,6 +275,24 @@ private void TestMvcMusicStore(string version)
             Assert.AreEqual(mvcProjectActions.Count, 4);
         }
 
+        [Test]
+        public void TestAntlrSampleSolution()
+        {
+            var version = "netcoreapp3.1";
```
Eruanion, line 281: Probably want to follow the example of the other tests and pull out the version to a separate function so we can inject the version number ourselves and run .net 5 if need be.

src/CTA.Rules.Models/ProjectActions.cs
Eruanion, line 17: Does this get used anywhere for now?

jonlouie: I was getting null reference errors for it. Not sure about other collections, but List properties aren't initialized or given a default value unless specified.

jonlouie: Yes, it is used elsewhere in the class but was not initialized.

tst/CTA.Rules.Test/PortCoreTests.cs
jonlouie: Good point. I will update this.

marknfawaz (approved):

PR #39, opened 2021-02-09
New actions and updated owin unit tests

Eruanion:
*Issue #, if available:*
https://github.com/aws/cta/issues/40

*Description of changes:*
New replace method modifiers action.
New add global expression action.
New remove constructor initializer action.
New add expression within constructor action.
New add expression within a method action.
Updated Owin Branching Pipelines unit test with all above new tests.
Updated all Owin unit tests with checks on version and packages.
Temp folder take precedence over S3 bucket rules.


src/CTA.Rules.Actions/ClassActions.cs
```diff
@@ -152,5 +151,100 @@ public class ClassActions
...
+                if (!parsedExpression.FullSpan.IsEmpty)
+                {
+                    var nodeDeclarations = node.Members;
+                    nodeDeclarations = nodeDeclarations.Insert(0,parsedExpression);
+                    node = node.WithMembers(nodeDeclarations).NormalizeWhitespace();
+                }
+                return node;
+            };
+            return RenameClass;
+        }
```
jonlouie, line 194: Two comments here:

1. Would it make sense to rename this function to `GetAddExpressionAction`, similar to what's in `MethodDeclarationActions.cs`? My thought is that `GlobalExpression` implies global scope, but this action permits the user to add expressions that are limited to just the class as well.

2. Should `RenameClass` be renamed to something like `AddGlobalExpression`?

src/CTA.Rules.Actions/MethodDeclarationActions.cs
```diff
@@ -23,5 +23,22 @@ public class MethodDeclarationActions
...
+                if(!statementExpression.FullSpan.IsEmpty)
+                {
+                    BlockSyntax nodeBody = node.Body;
+                    nodeBody = nodeBody.AddStatements(statementExpression);
+                    node = node.WithBody(nodeBody).NormalizeWhitespace();
+                }
+                return node;
+            };
+            return AddExpression;
+        }
```
jonlouie, line 42: Since this adds an expression to the end of a function, would it make sense to rename this action to `AppendExpression`? If we have other actions that behave similarly, then I think it's better to stick with the convention. If not, renaming might make sense since implementing the ability to specify where to insert an expression is a non-trivial task.

src/CTA.Rules.Actions/ClassActions.cs
Eruanion: 1. Agreed, renamed as mentioned.
2. I would argue that rename class is straightforward and would prefer to leave it as is. Mark checked in that action.

src/CTA.Rules.Actions/MethodDeclarationActions.cs
Eruanion: Agreed, I went ahead and renamed it and also renamed the class action AddConstructorExpression to AppendConstructorExpression also to follow this schema.

jonlouie (approved):
LGTM

jonlouie:
Looks like there are some merge conflicts that will need to be resolved. Can re-review after those changes are committed

jonlouie (approved):

marknfawaz (approved):

PR #41, opened 2021-02-10
Add actions to migrate code that reads configuration

### LINKED ISSUE #30: Add rules for migrating configuration read settings
When migrating settings from web.config to appSettings.json, the code that reads these settings will stop working. We need to check if we can also port the code that reads these settings. For example:

ConfigurationManager.ConnectionStrings["keyName"].ConnectionString -> ConfigurationManager.Configuration["keyName"]
AND 
WebConfigurationManager.AppSettings["stringUtilSetting"] -> ConfigurationManager.Configuration["stringUtilSetting"]

marknfawaz:
*Issue #, if available:*
Closes #30

*Description of changes:*

### DIFF
--- a/src/CTA.Rules.Analysis/RulesAnalysis.cs
+++ b/src/CTA.Rules.Analysis/RulesAnalysis.cs
@@ -194,6 +194,45 @@ namespace CTA.Rules.Analyzer
                                 if (AnalyzeChildren(fileAction, child.Children, ++level, parentNamespace, parentClass)) { containsActions = true; }
                                 break;
                             }
+                        case IdConstants.ElementAccessIdName:
+                            {
+                                ElementAccess elementAccess = (ElementAccess)child;
+                                var compareToken = new ElementAccessToken()
+                                {
+                                    Key = elementAccess.Expression,
+                                    FullKey = GetFullKey(elementAccess.Reference?.Namespace, elementAccess.SemanticClassType, elementAccess.Expression),
+                                    Type = elementAccess.SemanticClassType,
+                                    Namespace = elementAccess.Reference?.Namespace
+                                };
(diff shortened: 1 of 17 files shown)


src/CTA.Rules.Analysis/RulesAnalysis.cs
```diff
@@ -246,6 +285,22 @@ private bool AnalyzeChildren(FileActions fileAction, UstList<UstNode> children,
...
+            }
+            else
+            {
+                if (!string.IsNullOrEmpty(containingClass))
+                {
+                    return $"{containingNamespace}.{containingClass}.{key}";
+                }
+                return $"{containingNamespace}.{key}";
+            }
+        }
```
jonlouie, line 302: Is it possible to have `$"{containingClass}.{key}"`? If so, one nit/suggestion is to simplify this by using fewer branching blocks:

```csharp
private string GetFullKey(string containingNamespace, string containingClass, string key)
{
	var namespaceToken = string.IsNullOrEmpty(containingNamespace) 
		? string.Empty 
		: $"{containingNamespace}.";

	var classToken = string.IsNullOrEmpty(containingClass) 
		? string.Empty 
		: $"{containingClass}.";

	return $"{namespaceToken}{classToken}{key}";
}
```

jonlouie (commented):
Only had one question/comment. Let me know if it makes sense, otherwise everything else lgtm

jonlouie (approved):

src/CTA.Rules.Analysis/RulesAnalysis.cs
marknfawaz: As discussed, we don't want to use containingClass.key because then the namespace would be empty (meaning semantic analysis failed)

jonlouie (approved):

PR #56, opened 2021-02-25
Object creation changes

Eruanion:
## Related issue

Closes: #57 

## Description
Added new actions to modify properties when creating new objects.
One for adding new properties.
One for replacing property identifiers.
Once for adding comments.
Updated Owin unit tests for static files.

## Supplemental testing
All current unit tests are passing.

## Additional context
https://sim.amazon.com/issues/53a138ab-0c1e-424d-a9ec-ca2bd1f49923

---

### DIFF
--- a/src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
+++ b/src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
@@ -40,29 +40,51 @@ namespace CTA.Rules.Actions
         {
             Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> action = (SyntaxGenerator syntaxGenerator, ObjectCreationExpressionSyntax node) =>
             {
-                var member = node.Initializer.Expressions.FirstOrDefault(n => n.ToFullString().Contains(oldMember));
-                if(member != null)
+                if(node.Initializer != null && node.Initializer.Expressions.Count > 0)
                 {
-                    var newExpression = SyntaxFactory.ParseExpression(member.ToFullString().Replace(oldMember,newMember));
-                    var newNode = node.Initializer.Expressions.Replace(member, newExpression);
-                    node = node.WithInitializer(node.Initializer.WithExpressions(newNode)).NormalizeWhitespace();
-                }
-                else
-                {
-                    var newExpression = SyntaxFactory.ParseExpression(newMember + "=" + newValueIfAdding);
-                    var newNode = node.Initializer.Expressions.Add(newExpression);
-                    node = node.WithInitializer(node.Initializer.WithExpressions(newNode)).NormalizeWhitespace();
+                    var memberList = node.Initializer.Expressions.Where(n => n.Kind() == SyntaxKind.SimpleAssignmentExpression);
+                    if(memberList.Count() > 0)
+                    {
+                        var assignMemberList = memberList.Select(n => (AssignmentExpressionSyntax)n);
+                        var member = assignMemberList.Where(n => n.Left.ToFullString().Contains(oldMember)).FirstOrDefault();
+                        if (member != null)
+                        {
+                            var newExpression = SyntaxFactory.ParseExpression(newMember + "=" + member.Right);
+                            var newNode = node.Initializer.Expressions.Replace(member, newExpression);
+                            node = node.WithInitializer(node.Initializer.WithExpressions(newNode)).NormalizeWhitespace();
+                        }
+                        else
+                        {
+                            var newExpression = SyntaxFactory.ParseExpression(newMember + "=" + newValueIfAdding);
+                            var newNode = node.Initializer.Expressions.Add(newExpression);
+                            node = node.WithInitializer(node.Initializer.WithExpressions(newNode)).NormalizeWhitespace();
+                        }
+                    }
                 }
                 return node;
             };
             return action;
         }
 
-        public Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> GetReplaceObjectIdentifierAction(string identifier)
+        public Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> GetReplaceObjectPropertyValueAction(string oldMember, string newMember)
         {
             Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> action = (SyntaxGenerator syntaxGenerator, ObjectCreationExpressionSyntax node) =>
             {
-                node = node.WithType(SyntaxFactory.ParseTypeName(identifier)).NormalizeWhitespace();
+                if (node.Initializer != null && node.Initializer.Expressions.Count > 0)
+                {
+                    var memberList = node.Initializer.Expressions.Where(n => n.Kind() == SyntaxKind.SimpleAssignmentExpression);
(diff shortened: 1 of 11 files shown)


src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
```diff
@@ -34,5 +36,71 @@ ExpressionSyntax action(SyntaxGenerator syntaxGenerator, ObjectCreationExpressio
...
+                if (node.Initializer != null && node.Initializer.Expressions.Count > 0)
+                {
+                    var memberList = node.Initializer.Expressions.Where(n => n.Kind() == SyntaxKind.SimpleAssignmentExpression);
+                    if (memberList.Count() > 0)
+                    {
+                        var assignMemberList = memberList.Select(n => (AssignmentExpressionSyntax)n);
+                        var member = assignMemberList.Where(n => n.Right.ToFullString().Contains(oldMember)).FirstOrDefault();
+                        if (member != null)
+                        {
+                            var newExpression = SyntaxFactory.ParseExpression(member.Left + "=" + member.Right.ToString().Replace(oldMember,newMember));
```
jonlouie, line 82: What do you think about having a helper class to help build certain types of nodes more easily? For example, here (and on lines 52 and 58 above) we're creating assignment expressions using an undefined pattern which could be substituted for something like `NodeFactory.CreateAssignmentExpression(leftOperand, rightOperand)`. 

I think making it easier to construct various node types would help us and the community w/ maintainability and testability, but I also don't think this is an urgent need. If we think there's value there, we can file an issue to add it later.

src/CTA.Rules.Update/ActionsRewriter.cs
```diff
@@ -321,11 +321,12 @@ public override SyntaxNode VisitCompilationUnit(CompilationUnitSyntax node)
         }
         public override SyntaxNode VisitObjectCreationExpression(ObjectCreationExpressionSyntax node)
         {
-            ObjectCreationExpressionSyntax newNode = node;// (ObjectCreationExpressionSyntax)base.VisitObjectCreationExpression(node);
-            bool skipChildren = false;
+            var symbols = _semanticModel.GetSymbolInfo(node);
+            ExpressionSyntax newNode = node;// (ObjectCreationExpressionSyntax)base.VisitObjectCreationExpression(node);
```
jonlouie, line 325: Can we remove the commented code here?

src/CTA.Rules.Update/ActionsRewriter.cs
```diff
@@ -321,11 +321,12 @@ public override SyntaxNode VisitCompilationUnit(CompilationUnitSyntax node)
...
+            bool skipChildren = false; // This is here to skip actions on children node when the main dientifier was changed. Just use new expression for the subsequent children actions.
```
jonlouie, line 326: Fix typo in "identifier"

jonlouie (changes requested):

src/CTA.Rules.Update/ActionsRewriter.cs
```diff
@@ -321,11 +321,12 @@ public override SyntaxNode VisitCompilationUnit(CompilationUnitSyntax node)
...
             foreach (var action in _fileActions.ObjectCreationExpressionActions)
             {
-                if (newNode.ToString() == action.Key)
+                if (newNode.ToString() == action.Key || symbols.Symbol.ContainingType.Name == action.Key)
```
marknfawaz, line 329: Need to add a null check here (in case symbol is null)

src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
```diff
@@ -34,5 +36,71 @@ ExpressionSyntax action(SyntaxGenerator syntaxGenerator, ObjectCreationExpressio
             return action;
         }
 
+        public Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> GetReplaceOrAddObjectPropertyIdentifierAction(string oldMember, string newMember, string newValueIfAdding)
+        {
+            Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> action = (SyntaxGenerator syntaxGenerator, ObjectCreationExpressionSyntax node) =>
```
marknfawaz, line 41: This declaration can be simplified. Please run code analysis on the commit to get suggestions from editorconfig.

src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
```diff
@@ -34,5 +36,71 @@ ExpressionSyntax action(SyntaxGenerator syntaxGenerator, ObjectCreationExpressio
...
+                    }
+                }
+                return node;
+            };
+            return action;
+        }
+
+        public Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> GetReplaceObjectPropertyValueAction(string oldMember, string newMember)
+        {
+            Func<SyntaxGenerator, ObjectCreationExpressionSyntax, ExpressionSyntax> action = (SyntaxGenerator syntaxGenerator, ObjectCreationExpressionSyntax node) =>
```
marknfawaz, line 71: This declaration can be simplified. Please run code analysis on the commit to get suggestions from editorconfig.

marknfawaz (changes requested):

src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
Eruanion: Good idea, Let's add a new issue for it, I don't want to significantly increase the amount of code to be re-reviewed for this CR and this idea can easily find other functions to refactor out of the other actions classes as well to include in this new helper class.

src/CTA.Rules.Update/ActionsRewriter.cs
Eruanion: Yep, I left it there only for context of what was there before but the CR file diff achieves that perfectly on its own. Removed as suggested.

src/CTA.Rules.Update/ActionsRewriter.cs
Eruanion: Thanks :-)

src/CTA.Rules.Update/ActionsRewriter.cs
Eruanion: Thanks Mark, good catch, I went ahead and used the '?' character after symbols.Symbol? to force a null return to avoid have a null reference exception and I added one to symbols.Symbol?.ContainingType? as well to be safe.

src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
Eruanion: Awesome, thanks Mark. I will always check messages from code analysis from now on. Completed as mentioned.

src/CTA.Rules.Actions/ObjectCreationExpressionActions.cs
Eruanion: Awesome, thanks Mark. I will always check messages from code analysis from now on. Completed as mentioned.

marknfawaz (approved):

jonlouie (approved):

Issue #66, opened 2021-03-05
Rule matches 2 Start method from different namespaces


Eruanion:
## Describe the bug
Currently Microsoft.Owin.Hosting.WebApp.Start and System.Diagnostics.Process.Start both match against an invocation expression rule for only the former.

### Expected behavior
The rule should only match against the first method from the Owin namespace.

### Steps to Reproduce
1. Execute the TestEmbedded test for either .net 3.1 or .net 5 and verify the output for the program.cs class.

### Desktop (please complete the following information)
 - OS: [e.g. Windows, macOS, Linux, etc]
 - Version/Distro: [e.g. 11, Big Sur, Ubuntu 18.04]

PR #67, opened 2021-03-05
Invocation too vague

Eruanion:
## Related issue

Closes: #66 

## Description
The invocation expressions keys are too imprecise causing false positive matches against methods of the same name in other namespaces. This makes the key more specific and prevents this from happening. 

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Analysis/RulesAnalysis.cs
+++ b/src/CTA.Rules.Analysis/RulesAnalysis.cs
@@ -268,9 +268,9 @@ namespace CTA.Rules.Analyzer
                                 }
 
                                 token = null;
-                                if(!string.IsNullOrEmpty(objectCreationNode.SemanticMethodSignature))
+                                if(!string.IsNullOrEmpty(objectCreationNode.SemanticOriginalDefinition))
                                 {
-                                    var nameToken = new ObjectCreationExpressionToken() { Key = objectCreationNode.SemanticClassType, Namespace = objectCreationNode.SemanticNamespace, Type = objectCreationNode.SemanticClassType };
+                                    var nameToken = new ObjectCreationExpressionToken() { Key = objectCreationNode.SemanticOriginalDefinition, Namespace = objectCreationNode.SemanticNamespace, Type = objectCreationNode.SemanticClassType };
                                     _rootNodes.ObjectCreationExpressionTokens.TryGetValue(nameToken, out token);
                                     if (token != null)
                                     {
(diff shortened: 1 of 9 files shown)


src/CTA.Rules.Analysis/RulesAnalysis.cs
```diff
@@ -184,7 +184,7 @@ private bool AnalyzeChildren(FileActions fileAction, UstList<UstNode> children,
                         case IdConstants.InvocationIdName:
                             {
                                 InvocationExpression invocationExpression = (InvocationExpression)child;
-                                var compareToken = new InvocationExpressionToken() { Key = invocationExpression.MethodName, Namespace = invocationExpression.Reference.Namespace, Type = invocationExpression.SemanticClassType };
+                                var compareToken = new InvocationExpressionToken() { Key = invocationExpression.SemanticOriginalDefinition, Namespace = invocationExpression.Reference.Namespace, Type = invocationExpression.SemanticClassType };
                                 _rootNodes.Invocationexpressiontokens.TryGetValue(compareToken, out var token);
                                 if (token != null)
                                 {
```
jonlouie, line 190: This looks like the right thing to do, but were any rules affected by this?

src/CTA.Rules.Update/ActionsRewriter.cs
```diff
@@ -195,27 +195,35 @@ public override SyntaxNode VisitInvocationExpression(InvocationExpressionSyntax
             var symbols = _semanticModel.GetSymbolInfo(node);
             var newNode = (InvocationExpressionSyntax)base.VisitInvocationExpression(node);
 
-            foreach (var action in _fileActions.InvocationExpressionActions)
+            var symbol = symbols.Symbol;
+
+            if (symbol != null)
             {
```
jonlouie, line 201: Optional: replace the `if` block with a guard statement to reduce nesting

Example: 
```csharp
if (someCondition)
{
    return newNode;
}

// do stuff down here

```

src/CTA.Rules.Analysis/RulesAnalysis.cs
Eruanion: Oh yep, I had to fix all the invocation expression rules. There were quite a few for Owin and a couple in the system.Web rules file. Unfortunately after doing some checking the only way to fix this problem was to make the rule from precise which necessitated changing all the rules unfortunately.

src/CTA.Rules.Update/ActionsRewriter.cs
Eruanion: Thanks, will follow this rule from now on.

src/CTA.Rules.Analysis/RulesAnalysis.cs
jonlouie: Got it, makes sense. If the unit tests are all still working then this should be ok

jonlouie (approved):

marknfawaz (approved):

Issue #69, opened 2021-03-10
Add detection for WindowsAuth types


jonlouie:
## Description
Add detection for various WindowsAuth types:
* WindowsAuthentication
* WindowsAuthorization
* WindowsAuthorizationRoles
* WindowsImpersonation

PR #70, opened 2021-03-10
Windows auth feature detection

jonlouie:
## Related issue

Closes: #69

## Description
* Add feature detection for the following windows auth types:
  * WindowsAuthentication
  * WindowsAuthorization
  * WindowsAuthorizationRoles
  * WindowsImpersonation

## Supplemental testing

## Additional context

---


smhmhmd (dismissed):

birojnayak:
why this is not in the path where CTA asses the config file for transformation and where we capture each API... let's discuss..

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WebConfigFeature.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Xml.Linq;
+using Codelyzer.Analysis;
+using CTA.FeatureDetection.Common.Models.Features.Base;
+using CTA.Rules.Config;
+
+namespace CTA.FeatureDetection.AuthType.CompiledFeatures
+{
+    public abstract class WebConfigFeature : CompiledFeature
+    {
+        private const string WebConfigFileName = Rules.Config.Constants.WebConfig;
```
Eruanion, line 14: Do we really need a local variable here for this constant? Since it is not being changed can we just reference it directly from the Constants class?

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WindowsAuthorizationRolesFeature.cs
```diff
@@ -0,0 +1,53 @@
+﻿using System.Linq;
+using Codelyzer.Analysis;
+using Codelyzer.Analysis.Model;
+using CTA.FeatureDetection.Common.Extensions;
+
+namespace CTA.FeatureDetection.AuthType.CompiledFeatures
+{
+    public class WindowsAuthorizationRolesFeature : WebConfigFeature
+    {
+        private readonly string _path = $"{Constants.ConfigurationElement}/{Constants.SystemWebElement}/{Constants.AuthorizationElement}/{Constants.AllowElement}";
```
Eruanion, line 10: We are computing local path variables often can we add these as part of the constants class instead to be put together from its own members?

Eruanion (changes requested):
Good work Jon, good clean code. Just a couple of comments included.

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WebConfigFeature.cs
jonlouie: Currently doing some refactoring- this will be removed

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WindowsAuthorizationRolesFeature.cs
jonlouie: Will do- thanks!

Eruanion (approved):

Eruanion (dismissed):

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WindowsAuthorizationFeature.cs
```diff
@@ -0,0 +1,46 @@
...
+            var config = LoadWebConfig(analyzerResult.ProjectResult.ProjectRootPath);
+            return config.ContainsElement(Constants.AuthorizationElementPath);
+        }
+
+        private bool IsPresentInCode(AnalyzerResult analyzerResult)
+        {
+            var allAttributes = analyzerResult.ProjectResult.SourceFileResults.SelectMany(r => r.AllAnnotations());
+            return allAttributes.Any(a => a.Identifier == Constants.AuthorizeMethodAttribute);
+        }
+    }
```
marknfawaz, line 45: Are we saying here the authorize attribute means we're using windows auth? Authorize can be used when authenticating with other providers.

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WindowsAuthorizationRolesFeature.cs
```diff
@@ -0,0 +1,50 @@
...
+            return config.ContainsAttribute(Constants.AllowElementPath, Constants.RolesAttribute);
+        }
+        
+        private bool IsPresentInCode(AnalyzerResult analyzerResult)
+        {
+            var allAttributes = analyzerResult.ProjectResult.SourceFileResults.SelectMany(r => r.AllAnnotations());
+            var authorizeAttributes = allAttributes.Where(a => a.Identifier == Constants.AuthorizeMethodAttribute);
+
+            return authorizeAttributes.SelectMany(a => a.AllAttributeArguments())
+                .Any(a => a.ArgumentName == Constants.RolesAttributeArgument);
```
marknfawaz, line 47: same comment as above. Authorize can be used with non-windows authentication

src/CTA.Rules.Common/WebConfigManagement/WebConfigManager.cs
```diff
@@ -0,0 +1,77 @@
...
+                }
+
+                var xDocument = XDocument.Load(webConfigFile);
+                XDocumentCache[webConfigFile] = xDocument;
+
+                return XDocumentCache[webConfigFile];
+            }) as XDocument;
+
+            return new WebConfigXDocument(config);
+        }
```
marknfawaz, line 57: Which load method are we using? Can we remove the other one. Also, are we clearing the cache after finishing the operation?

tst/CTA.Rules.Test/CTA.Rules.Test.csproj
```diff
@@ -30,6 +31,12 @@
     <None Update="CTAFiles\action.namespace.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Update="CTAFiles\Configs\Web.config">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Update="CTAFiles\Configs\authWeb.config">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
```
marknfawaz, line 39: I suggest adding to a different folder to avoid later issues. CTA files should be used for rule files only.

marknfawaz (changes requested):

src/CTA.FeatureDetection.AuthType/CompiledFeatures/FormsAuthenticationWithMembershipFeature.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// 
+        /// </summary>
+        /// <param name="analyzerResult">Source code analysis results</param>
+        /// <returns>Whether or not Forms Authentication with Membership is used</returns>
+        public override bool IsPresent(AnalyzerResult analyzerResult)
+        {
+            var config = WebConfigManager.LoadWebConfigAsXDocument(analyzerResult.ProjectResult.ProjectRootPath);
+
+            return config.ContainsElement(Constants.MembershipElementPath)
+                   && config.ContainsAttributeWithValue(Constants.AuthenticationElementElementPath, Constants.ModeAttribute, Constants.FormsAuthenticationType);
```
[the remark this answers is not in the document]
jonlouie: Yes, I can make this a child class and change the logic to:
```csharp
return base.IsPresent(...) && config.ContainsElement(Constants.MembershipElementPath);
```

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WindowsAuthorizationRolesFeature.cs
jonlouie: same as above

src/CTA.Rules.Common/WebConfigManagement/WebConfigManager.cs
jonlouie: And currently we are not handling cache clearing- I'll include that as well.

src/CTA.FeatureDetection.AuthType/CompiledFeatures/WindowsAuthorizationFeature.cs
jonlouie: @marknfawaz Seems that you are correct. [This thread](https://stackoverflow.com/questions/1385042/asp-net-mvc-forms-authentication-authorize-attribute-simple-roles) implies that `[Authorize]` will use any underlying provider (in that example, Forms Auth).

src/CTA.Rules.Common/WebConfigManagement/WebConfigManager.cs
```diff
@@ -0,0 +1,83 @@
...
+                return _xDocumentCache[webConfigFile];
+            }) as XDocument;
+
+            return new WebConfigXDocument(config);
+        }
+
+        public static void ClearCache()
+        {
+            _configurationCache = new Dictionary<string, Configuration>();
+            _xDocumentCache = new Dictionary<string, XDocument>();
```
marknfawaz, line 62: nit: Why not just set to null?

marknfawaz (changes requested):

src/CTA.Rules.Common/WebConfigManagement/WebConfigManager.cs
jonlouie: The cache objects are static so if we set them to null and try loading another config, we'll get a null reference error. Open to other ideas if there's a better way to approach this.

marknfawaz: One way to do it can be to have a get where the object is initialized if it's null. This way we're not creating an object unless we need it.

marknfawaz (changes requested):

src/CTA.Rules.Common/WebConfigManagement/WebConfigManager.cs
jonlouie: Sounds good, will add this

marknfawaz (approved):

Issue #83, opened 2021-03-21
Add feature to detect if a project is a core project


marknfawaz:
## Description
Add a feature to detect project type (core or framework)

## Solution idea(s)
In the list of features in a project, there should be a Core/Framework feature

PR #87, opened 2021-03-24
add DotnetFramework and DotnetCore features

jonlouie:
## Related issue

Closes: #83

## Description
* Detects if a project is .NET Framework, .NET Core, or .NET Standard
* Looks at `.csproj` files and uses regex patterns to identify the target framework type
* Since regex patterns are used, the project type is identified based on intent of the developer, not based on correctness (e.g. `netcoreapp.3.9` is an invalid target framework, but it would still register as a .NET Core project)

## Supplemental testing

### Sources of Target Framework patterns
* [Valid TargetFramework values](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks) used in SDK-style `.csproj` files
* [Valid TargetFrameworkVersion values](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-target-framework-and-target-platform?view=vs-2019#target-framework-and-profile:~:text=The%20available%20values%20for%20TargetFrameworkVersion%20are,v4.6.2%2C%20v4.7%2C%20v4.7.1%2C%20v4.7.2%2C%20and%20v4.8.) used in .NET Framework `.csproj` files

---


src/CTA.Rules.Common/Extensions/XDocumentExtensions.cs
```diff
@@ -20,5 +22,10 @@ public static bool ContainsAttributeValue(this XDocument document, string elemen
         {
             return document.XPathSelectElements(elementPath).Any(e => e.Attribute(attributeName)?.Value == value);
         }
+
+        public static IEnumerable<XElement> GetElementsByLocalName(this XDocument document, string elementName)
+        {
+            return document.Descendants().Where(d => string.Equals(d.Name.LocalName, elementName, StringComparison.InvariantCulture));
```
Eruanion, line 28: What if the document doesn't have any Descendants? Empty document? Would this still behave as intended through the stack trace?

Eruanion (commented):
Awesome code, just that one comment based on empty document possible problem.

src/CTA.Rules.Common/Extensions/XDocumentExtensions.cs
jonlouie: Good questions. So not having any descendants is the same as have an empty document, in which case this method would yield an empty enumerable which is the expected behavior. In the case of having a single element, `document.Descendants()` would yield an enumerable with just that one element.

Eruanion (approved):

src/CTA.Rules.Common/CsprojManagement/CsprojXDocument.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            return IsRegexMatch(Constants.DotnetStandardPattern, TargetFrameworksValue);
+        }
+
+        private string GetTargetFrameworksValue()
+        {
+            var targetFrameworks =
+                _csproj.GetElementsByLocalName(Constants.TargetFrameworkVersionElement).FirstOrDefault()
+                ?? _csproj.GetElementsByLocalName(Constants.TargetFrameworkElement).FirstOrDefault()
+                ?? _csproj.GetElementsByLocalName(Constants.TargetFrameworksElement).FirstOrDefault();
```
marknfawaz, line 53: Can both TargetFramework and TargetFrameworks exist at the same time? If yes, this will not give a correct result

marknfawaz (changes requested):
One more condition to consider is the global.json file that sets the version for all projects within a directory:

https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x

src/CTA.Rules.Common/CsprojManagement/CsprojXDocument.cs
jonlouie: Yes, it is technically valid for both `TargetFramework` and `TargetFrameworks` to exist in a .csproj file, but in those cases, `TargetFrameworks` will be ignored ([source](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#targetframeworks:~:text=This%20property%20is%20ignored%20if%20TargetFramework%20(singular)%20is%20specified.))

jonlouie:
> One more condition to consider is the global.json file that sets the version for all projects within a directory:
> 
> https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x

@marknfawaz This is an interesting case. At first glance, it doesn't seem like global.json is coupled to a project and only customizes the CLI execution environment (as opposed to declaring the target framework inherent to the project). I don't think this is a case we'll need to factor in here, but I could be wrong. If my understanding is incorrect, we can talk discuss more in detail.

jonlouie:
> > One more condition to consider is the global.json file that sets the version for all projects within a directory:
> > https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x
> 
> @marknfawaz This is an interesting case. At first glance, it doesn't seem like global.json is coupled to a project and only customizes the CLI execution environment (as opposed to declaring the target framework inherent to the project). I don't think this is a case we'll need to factor in here, but I could be wrong. If my understanding is incorrect, we can talk discuss more in detail.

Per our conversation, I will update the features to use the build result to get the target framework(s).

Eruanion (approved):

saikiranakula-amzn (approved):

jonlouie:
Closing- obsolete

Issue #100, opened 2021-04-06
Update feature detection config files used for testing


jonlouie:
## Description of issue
Config files should use more precise pattern matching

## Additional context
Provide any additional information

PR #101, opened 2021-04-06
update feature config pattern matching

jonlouie:
## Related issue

Closes: #100

## Description
* Update test feature configs to use more precise matching
* The updated file name patterns now only match file paths that end with the specified character string
* The test files will match what is used in production

## Supplemental testing

## Additional context

---

### DIFF
--- a/tst/CTA.FeatureDetection.Tests/Examples/Templates/EntityFramework_config_features.json
+++ b/tst/CTA.FeatureDetection.Tests/Examples/Templates/EntityFramework_config_features.json
@@ -9,7 +9,7 @@
           "Condition": {
             "Type": "XmlFileQuery",
             "Properties": {
-              "FileNamePattern": "App.config|Web.config",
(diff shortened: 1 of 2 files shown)


Eruanion (approved):

marknfawaz (changes requested):
I'm not sure I understand this change.. Is there a code update somewhere to go along with this, or will the code now automatically work with this?

jonlouie:
> I'm not sure I understand this change.. Is there a code update somewhere to go along with this, or will the code now automatically work with this?

The code will automatically work with this (`XmlFileQuery` has always matched FileNamePattern via regex). These changes will just allow us to match test files with what is currently in S3 (S3 file had to be updated to prevent inadvertent files matches, e.g. `~/some/path/app/config.cs` would match when targeting `app.confg`)

marknfawaz (approved):

Issue #13, opened 2021-01-08
Action validation improvements


marknfawaz:
The way actions are validated has to be enhanced to be more robust. We can change it to use regular expressions, or count the occurrences of the validation string before and after

PR #102, opened 2021-04-13
Action validations

Eruanion:
## Related issue

Closes: #13 

## Description
Added a new action validation field for checking comments, otherwise a regex is run to remove all strings and comments from final files to check action validations. Updates all relevant rules file to include this new check if needed. Also removed a bug with ActionRewriters which was doubling all actions.

## Supplemental testing
All unit tests were checked for validity of actions and all passed.

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Models/Actions/ActionValidation.cs
+++ b/src/CTA.Rules.Models/Actions/ActionValidation.cs
@@ -4,5 +4,6 @@
     {
         public string Contains { get; set; }
         public string NotContains { get; set; }
+        public string CheckComments { get; set; }
     }
 }
--- a/src/CTA.Rules.Update/ActionsRewriter.cs
+++ b/src/CTA.Rules.Update/ActionsRewriter.cs
@@ -357,7 +357,6 @@ namespace CTA.Rules.Update.Rewriters
                         actionExecution.InvalidExecutions = 1;
                         LogHelper.LogError(actionExecutionException);
                     }
-                    allActions.Add(actionExecution);
                 }
             }
             if (!skipChildren)
--- a/src/CTA.Rules.Update/CodeReplacer.cs
+++ b/src/CTA.Rules.Update/CodeReplacer.cs
@@ -3,8 +3,10 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using Codelyzer.Analysis.Build;
+using CTA.FeatureDetection.Common.Extensions;
 using CTA.Rules.Config;
 using CTA.Rules.Models;
 using CTA.Rules.Update.Rewriters;
@@ -161,14 +163,23 @@ namespace CTA.Rules.Update
 
         public List<GenericActionExecution> ValidateActions(List<GenericActionExecution> actions, string fileResult)
         {
-            var trimmedResult = Utils.EscapeAllWhitespace(fileResult);
+            // Matches all types of comments and strings
+            string regComments = @"\/\*(?:(?!\*\/)(?:.|[\r\n]+))*\*\/|\/\/(.*?)\r?\n|""((\\[^\n]|[^""\n])*)""|@(""[^""""]*"")+";
 
             foreach (var action in actions)
             {
                 var actionValidation = action.ActionValidation;
+                var trimmedResult = fileResult;
                 var actionValid = true;
 
                 if (actionValidation == null) { continue; }
+
+                if (string.IsNullOrEmpty(actionValidation.CheckComments) || !bool.Parse(actionValidation.CheckComments))
+                {
+                    trimmedResult = Regex.Replace(fileResult, regComments, "");
+                }
+
+                trimmedResult = Utils.EscapeAllWhitespace(trimmedResult);
 
                 var contains = !string.IsNullOrEmpty(actionValidation.Contains) ? Utils.EscapeAllWhitespace(actionValidation.Contains) : "";
                 var notContains = !string.IsNullOrEmpty(actionValidation.NotContains) ? Utils.EscapeAllWhitespace(actionValidation.NotContains) : "";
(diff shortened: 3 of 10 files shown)


ShanshanHe (dismissed):
lgtm

src/CTA.Rules.Models/Actions/ActionValidation.cs
```diff
@@ -4,5 +4,6 @@ public class ActionValidation
     {
         public string Contains { get; set; }
         public string NotContains { get; set; }
+        public string CheckComments { get; set; }
```
jonlouie, line 7: Can this be changed to a bool? JSON format permits boolean values (and bool properties default to `false` which I believe is the intended behavior here).
Example:
```
{
    "boolField1": true,
    "boolField2": false
}
```

src/CTA.Rules.Update/CodeReplacer.cs
```diff
@@ -161,15 +163,24 @@ private List<GenericActionExecution> AddActionsWithoutExecutions(FileActions cur
...
 
             foreach (var action in actions)
             {
                 var actionValidation = action.ActionValidation;
+                var trimmedResult = fileResult;
                 var actionValid = true;
 
                 if (actionValidation == null) { continue; }
 
+                if (string.IsNullOrEmpty(actionValidation.CheckComments) || !bool.Parse(actionValidation.CheckComments))
```
jonlouie, line 177: Changing `CheckComments` to a bool type would simplify this `if` statement as well

tst/CTA.Rules.Test/TempRules/microsoft.owin.hosting.json
```diff
@@ -115,7 +115,8 @@
               "Description": "Add a comment to explain how to replace WebApp.Start.",
               "ActionValidation": {
                 "Contains": "ReplaceMicrosoft.Owin.Hosting.WebApp.StartwithWebHostBuilder",
-                "NotContains": ""
+                "NotContains": "",
+                "CheckComments":  "True"
```
jonlouie, line 119: If changing `CheckComments` to bool, need to update value to `true` (this applies to other occurrences below).

src/CTA.Rules.Update/CodeReplacer.cs
```diff
@@ -161,15 +163,24 @@ private List<GenericActionExecution> AddActionsWithoutExecutions(FileActions cur
 
         public List<GenericActionExecution> ValidateActions(List<GenericActionExecution> actions, string fileResult)
         {
-            var trimmedResult = Utils.EscapeAllWhitespace(fileResult);
+            // Matches all types of comments and strings
+            string regComments = @"\/\*(?:(?!\*\/)(?:.|[\r\n]+))*\*\/|\/\/(.*?)\r?\n|""((\\[^\n]|[^""\n])*)""|@(""[^""""]*"")+";
```
marknfawaz, line 167: Does this work for other types of comments? Example:
`// This is a comment`
or
`/// <summary>`
`/// Initializes a new instance of Solution Port, analyzing the solution path using the provided config.`
`/// </summary>`

I tried using a regex testing side and it doesn't seem to recognize them: http://regexstorm.net/tester

Can we load a syntax tree without comments? One example I was able to run was creating an extension method that loads the children without their trivia. This will ensure that we are using Roslyn APIs to get rid of comments instead of recreating our own:


        public static SyntaxNode NoComments(this SyntaxNode node)
        {
            if (node != null)
            {
                var childNodes = node.ChildNodes();
                node = node.ReplaceNodes(childNodes, RemoveComments);
                node = node.WithoutTrivia();
            }
            return node;
        }
        static SyntaxNode RemoveComments(SyntaxNode syntaxNode, SyntaxNode returnNode)
        {
            return syntaxNode.NoComments();
        }

marknfawaz (dismissed):

src/CTA.Rules.Update/ActionsRewriter.cs
```diff
@@ -357,7 +357,6 @@ public override SyntaxNode VisitObjectCreationExpression(ObjectCreationExpressio
                         actionExecution.InvalidExecutions = 1;
                         LogHelper.LogError(actionExecutionException);
                     }
-                    allActions.Add(actionExecution);
```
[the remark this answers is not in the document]
Eruanion: Yeah this is present twice, a few lines above inside the try catch statement also. This is causing loading of object expressions actions to happen twice. Which caused problems with the counting method that we discontinued but I still wanted to remedy this bug even though its current impact appears to be small currently.

tst/CTA.Rules.Test/TempRules/microsoft.owin.hosting.json
Eruanion: Mark also mentioned he would rather keep everything as strings for now and not introduce new data types currently.

src/CTA.Rules.Update/CodeReplacer.cs
Eruanion: Yup it works on all types of comments, I used : https://regexr.com/ but your site also worked.
Please use:
\/\*(?:(?!\*\/)(?:.|[\r\n]+))*\*\/|\/\/(.*?)\r?\n|"((\\[^\n]|[^""\n])*)"|@("[^"]*")+
For the pattern and you can use any variety of comments to test against, I find the below input is pretty thorough:

// remove whole line comments
bool broken = false; // remove partial line comments
if (broken == true)
{
    return "BROKEN";
}

/// This is another comment

/* This // is a test/, or // This / is a test */
var comment="/* This is a test*/";
var url = "http://stackoverflow.com";
var abc = @" this /* \n is a comment in quote\n*/ with or without spaces between ";
var abc2 = "and /*or*/ and ";

/* remove block comments
else
{
    return "FIXED";
} // do not remove nested comments */ bool working = !broken;
return "NO COMMENT";

src/CTA.Rules.Update/CodeReplacer.cs
jonlouie: If keeping it as a string, this looks ok to me 👍

src/CTA.Rules.Update/CodeReplacer.cs
Eruanion: I updated CR with the extensions all tests are still passing and comments are getting removed as intended. The only potential problem is if in the future we need to introduce more exceptions than just comments such as strings or others. I will keep the regex code commented out for potential re-introduction in the future if needed. I do think the extension idea is much cleaner however.

jonlouie (approved):

jonlouie (approved):

Issue #109, opened 2021-04-23
Enable incremental and parallel assessments


marknfawaz:
## Description
When porting to .net core, we need to make a few changes to enable file level assessment:
- Add the list of framework references to the ported project csproj file
- Create an interface to enable file level analysis

In addition, we need to return a list of textchanges to be able to preview changes before applying them.

PR #110, opened 2021-04-23
Incremental and porting updates

marknfawaz:
## Related issue

Closes: #109

### DIFF
--- a/src/CTA.Rules.Models/IDEFileActions.cs
+++ b/src/CTA.Rules.Models/IDEFileActions.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Codelyzer.Analysis.Model;
+
+namespace CTA.Rules.Models
+{
+    public class IDEFileActions
+    {
+        public string FilePath { get; set; }
+        public TextSpan TextSpan { get; set; }
+        public string Description { get; set; }
+        public IList<TextChange> TextChanges { get; set; }
+    }
+}
--- a/src/CTA.Rules.Models/TextChange.cs
+++ b/src/CTA.Rules.Models/TextChange.cs
@@ -1,0 +1,29 @@
+﻿using System;
+using Microsoft.CodeAnalysis;
+
+namespace CTA.Rules.Models
+{
+    public class TextChange
+    {
+        public string NewText { get; set; }
+        public FileLinePositionSpan FileLinePositionSpan { get; set; }
+
+        public bool Equals(TextChange textChange)
+        {
+            if (textChange == null) return false;
+
+            return NewText == textChange.NewText
+                && FileLinePositionSpan.Equals(textChange.FileLinePositionSpan);
+        }
+
+        public override bool Equals(object obj)
+        {
+            return Equals(obj as TextChange);
+        }
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(FileLinePositionSpan, NewText);
+        }
+    }
+}
--- a/src/CTA.Rules.Models/Tokens/NodeToken.cs
+++ b/src/CTA.Rules.Models/Tokens/NodeToken.cs
@@ -1,6 +1,7 @@
-﻿using System.Collections.Generic;
-using Codelyzer.Analysis.Model;
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
 using CTA.Rules.Config;
+using TextSpan = Codelyzer.Analysis.Model.TextSpan;
 
 namespace CTA.Rules.Models.Tokens
 {
@@ -24,6 +25,7 @@ namespace CTA.Rules.Models.Tokens
             ProjectLevelActions = new List<ProjectLevelAction>();
             ProjectFileActions = new List<ProjectLevelAction>();
             TargetCPU = new List<string>();
+            TextChanges = new List<TextChange>();
         }
         public string Key { get; set; }
         public string TrimmedKey => !string.IsNullOrEmpty(Key) ? Utils.EscapeAllWhitespace(Key) : string.Empty;
@@ -32,6 +34,7 @@ namespace CTA.Rules.Models.Tokens
         public string FullKey { get; set; }
         public TextSpan TextSpan { get; set; }
         public string Description { get; set; }
+        public IList<TextChange> TextChanges { get; set; }
         public List<string> TargetCPU { get; set; }
         public List<AttributeAction> AttributeActions { get; set; }
         public List<AttributeAction> AttributeListActions { get; set; }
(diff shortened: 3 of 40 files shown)


src/CTA.Rules.Models/IDEFileActions.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Codelyzer.Analysis.Model;
+
+namespace CTA.Rules.Models
+{
+    public class IDEFileActions
```
jonlouie, line 8: Could you add a doc comment for the class that describes its use and how it differs from other file actions?

Also, would it make sense to rename anything with the prefix `IDE` to something more generic? Dependencies shouldn't be aware of the libraries or projects that consume them, but I also don't have full context here about how this model is consumed.

src/CTA.Rules.ProjectFile/ProjectFileCreator.cs
```diff
@@ -115,14 +145,18 @@ public bool Create()
...
         {
-            if (string.IsNullOrEmpty(content))
+            if (string.IsNullOrEmpty(content?.Trim()))
             {
                 return string.Empty;
             }
-            
-            var itemGroupContent = string.Format(_itemGroupTemplate, content);
+
+            var currentGroupTemplate = itemGroupTemplate ?? _itemGroupTemplate;
```
jonlouie, line 157: nit: rename `_itemGroupTemplate` to `_defaultItemGroupTemplate`

src/CTA.Rules.Update/CodeReplacer.cs
```diff
@@ -44,23 +53,57 @@ public CodeReplacer(List<SourceFileBuildResult> sourceFileBuildResults, ProjectC
...
+                                    {
+                                        var fileLinePositionSpan = root.SyntaxTree.GetMappedLineSpan(textChange.Span);
+                                        var newTextChange = new TextChange() { FileLinePositionSpan = fileLinePositionSpan, NewText = textChange.NewText };
+                                        if (!nodetoken.TextChanges.Contains(newTextChange))
+                                        {
+                                            nodetoken.TextChanges.Add(newTextChange);
+                                        }
+                                    }
+                                });
+                            }
```
jonlouie, line 103: Each branch in this `if...else` block is quite large. What do you think about moving the logic for each branch into descriptive private functions to help with readability? For example:
```csharp
if (_projectConfiguration.PortCode)
{
    ApplyCodeActionsAndPersist( ... );
}
else
{
    // assumes TextChanges are used to preview changes
    ApplyCodeActionsAndPreview( ... ); 
}
```

src/CTA.Rules.Update/CodeReplacer.cs
```diff
@@ -72,56 +115,63 @@ public CodeReplacer(List<SourceFileBuildResult> sourceFileBuildResults, ProjectC
...
-                if (!string.IsNullOrEmpty(runResult))
+
+                if (!actionsPerProject.TryAdd(Constants.Project, projectRunActions))
                 {
-                    projectActionExecution.Description = string.Concat(projectActionExecution.Description, ": ", runResult);
-                    projectRunActions.Add(projectActionExecution);
-                    LogHelper.LogInformation(projectLevelAction.Description);
+                    LogHelper.LogError(new FilePortingException(Constants.Project, new Exception("Error adding project to actions collection")));
                 }
             }
```
jonlouie, line 174: Same as above- consider moving the logic in this `if` block to a private function called `ApplyProjectLevelActions` or something similar

src/CTA.Rules.Models/TextChange.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using Microsoft.CodeAnalysis;
+
+namespace CTA.Rules.Models
+{
+    public class TextChange
```
jonlouie, line 6: A doc comment for this class would be helpful as well

jonlouie (changes requested):

src/CTA.Rules.Models/Tokens/NodeToken.cs
```diff
@@ -51,5 +54,27 @@ public NodeToken()
...
+                allActions.AddRange(MemberAccessActions);
+                allActions.AddRange(IdentifierNameActions);
+                allActions.AddRange(InvocationExpressionActions);
+                allActions.AddRange(MemberAccessActions);
+                allActions.AddRange(UsingActions);
+                allActions.AddRange(ObjectCreationExpressionActions);
+                allActions.AddRange(NamespaceActions);
+                return allActions;
+            }
+        }
```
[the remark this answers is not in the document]
marknfawaz: I'm not sure I understand. How would this cause a memory leak?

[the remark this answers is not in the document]
jonlouie: Memory leak might not be the best term here, but my thought is that it's possible to inadvertently consume more resources than expected. See the below example for what I was thinking, but if this is a nonconcern, we can forgo this modification.

Consider this trivial example:
``` csharp
var allActions1 = nodeToken.AllActions;
var allActions2 = nodeToken.AllActions;
```
With the current implementation, this would allocate 2x the memory than the suggested implementation:
```csharp
private List<GenericAction> _allActions;
public List<GenericAction> AllActions
{
    get
    {
        _allActions = new List<GenericAction>();
        _allActions.AddRange(AttributeActions);
        _allActions.AddRange(AttributeListActions);
        _allActions.AddRange(MethodDeclarationActions);
        _allActions.AddRange(ClassDeclarationActions);
        _allActions.AddRange(InterfaceDeclarationActions);
        _allActions.AddRange(ElementAccessActions);
        _allActions.AddRange(MemberAccessActions);
        _allActions.AddRange(IdentifierNameActions);
        _allActions.AddRange(InvocationExpressionActions);
        _allActions.AddRange(MemberAccessActions);
        _allActions.AddRange(UsingActions);
        _allActions.AddRange(ObjectCreationExpressionActions);
        _allActions.AddRange(NamespaceActions);
        return _allActions;
    }
}
```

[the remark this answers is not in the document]
marknfawaz: As discussed, this wouldn't allocate 2x memory because it's a local variable that gets disposed after the getter exits. We can look into this and the same behavior in fileactions to see if there's a better way to do it.

src/CTA.Rules.ProjectFile/ProjectFileCreator.cs
```diff
@@ -44,26 +48,43 @@ public class ProjectFileCreator
...
 @"<ItemGroup>
 {0}
+</ItemGroup>";
+
+        private readonly string _portingInfoItemGroupTemplate =
+@"<ItemGroup Label=""PortingInfo"">
+<!-- DO NOT REMOVE WHILE PORTING
+{0}
+-->
 </ItemGroup>";
```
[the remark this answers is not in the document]
marknfawaz: Good idea. Please create an issue for this, because it'll require a change here and on codelyzer. we will prioritize this based on availability.

src/CTA.Rules.Models/IDEFileActions.cs
marknfawaz: We don't usually add doc comments for models. We can document our models classes at a later stage if needed.

huawenm (approved):

huawenm (approved):

src/CTA.Rules.ProjectFile/ProjectFileCreator.cs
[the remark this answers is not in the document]
jonlouie: Sure, will create those issues

jonlouie (approved):

Issue #123, opened 2021-06-03
Run translation on a project without using msbuild


marknfawaz:
## Description of issue
When the references for a project are available, we can get a semantic model without the need to run msbuild. This can be used to optimize the translation performance

PR #124, opened 2021-06-03
Feature detection and reference build updates

### LINKED ISSUE #122: Optimize the file lookup in feature detection
## Description of issue
When looking for files, instead of enumerating the whole folder multiple times then finding file patterns, we can run through this once and look only for specific patterns.

marknfawaz:
Update feature detection to avoid unnecessary loops. Add reference builds as an option when running CTA

## Related issue

Closes: #123 
Closes #122


Eruanion (dismissed):

jonlouie (approved):
As discussed, we will provide an update in the future to optimize our concurrency resolution

Issue #127, opened 2021-06-08
Report generator throws an error when run concurrently


jonlouie:
## Describe the bug
When reports are generated concurrently for the same project, an access error is thrown:
`The process cannot access the file 'C:\Some\Path\PortSolutionResult.txt' because it is being used by another process`

PR #128, opened 2021-06-08
create unique report files

jonlouie:
## Related issue

Closes: #127

## Description
* Reports are written to unique file names using ticks as a modifier. This prevents file access errors if the processes share the same file or if the user has a previous report open.

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Config/Utils.cs
+++ b/src/CTA.Rules.Config/Utils.cs
@@ -13,6 +13,8 @@ namespace CTA.Rules.Config
 {
     public class Utils
     {
+        private const string DefaultMutexName = "DefaultMutex";
+
         public static int GenerateHashCode(int prime, string str)
         {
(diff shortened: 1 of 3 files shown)


Eruanion (commented):
I would also prefer to create a single file per solution assessed and overwrite it upon subsequent assessment if possible.

jonlouie:
@marknfawaz @Eruanion Good suggestions. I updated the report generator to use a thread-safe write method. The write method allows file sharing across processes within the application. If there's an IOException on write, we know it is caused by an external lock on the file so a unique timestamp is appended to the file name. This allows new report files to be created, but only under the rare circumstance where the original file is locked by another application. If there is no external lock, the file is overwritten.

Eruanion (approved):

marknfawaz (approved):

PR #136, opened 2021-06-11
Added WorkspaceManagerService and Moved/Renamed Some Files

comsemer:
## Description
* Added WorkspaceManagerService and accompanying tests for workspace management functions and the status tracking and waiting functionality
* Also made some modifications to the project builder and analyzer classes which were then renamed and placed into the ProjectManagement folder
* Added some tests for the ProjectAnalyzer class changes
* Moved factory tests into factories folder in test project

---


tst/CTA.WebForms2Blazor.Tests/ProjectManagement/PartialProjectSetupFixture.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class PartialProjectSetupFixture
+    {
+        public const string TEST_AREA_DIRECTORY_NAME = "TestingArea";
+        public const string TEST_FILES_DIRECTORY_NAME = "TestFiles";
+        public const string TEST_STRUCTURE_1_DIRECTORY_NAME = "TestStructure1";
+        public const string TEST_BLAZOR_PROJECT_DIRECTORY_NAME = "BlazorTestProject";
+        public const string TEST_CLASS_FILE_NAME = "TestClassFile.cs";
+        public const string TEST_XML_FILE_NAME = "TestXMLFile.xml";
+        public const string NESTED_TEST_CLASS_FILE_NAME = "NestedTestClassFile.cs";
+        public const string NESTED_TEST_TEXT_FILE_NAME = "NestedTestTextFile.txt";
```
jonlouie, line 17: nit: use pascal case for constants to remain consistent with rest of projects in CTA

src/CTA.WebForms2Blazor/Services/WorkspaceManagerService.cs
```diff
@@ -0,0 +1,153 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Text;
+
+namespace CTA.WebForms2Blazor.Services
+{
+    public class WorkspaceManagerService
+    {
+        private const string WORKSPACE_DUPLICATE_ERROR = "Attempted to create a new Blazor workspace when one already exists";
+        private const string WORKSPACE_MISSING_ERROR_TEMPLATE = "Workspace {0} operation attempted, but no workspace exists";
+        private const string WORKSPACE_UPDATE_FAILED_ERROR_TEMPLATE = "Workspace {0} operation attempted, but workspace failed to apply changes";
```
jonlouie, line 15: nit: use pascal case for constants to remain consistent with rest of projects in CTA

src/CTA.WebForms2Blazor/Services/WorkspaceManagerService.cs
```diff
@@ -0,0 +1,153 @@
...
+            while (_numDocuments < _expectedDocuments)
+            {
+                if (token.IsCancellationRequested)
+                {
+                    throw new OperationCanceledException(token);
+                }
+
+                await Task.Delay(25);
+            }
+        }
```
jonlouie, line 151: Is there a reason for the explicit delays? Or is it just a placeholder so the compiler doesn't complain about the methods being `async` with no `await`?

src/CTA.WebForms2Blazor/ProjectManagement/ProjectAnalyzer.cs
```diff
@@ -1,23 +1,23 @@
...
         {
             _inputProjectPath = inputProjectPath;
         }
 
         public IEnumerable<FileInfo> GetProjectFileInfo()
         {
-            throw new NotImplementedException();
+            var directoryInfo = new DirectoryInfo(_inputProjectPath);
+            return directoryInfo.GetFiles("*", SearchOption.AllDirectories);
         }
```
jonlouie, line 21: Have you considered the case where we encounter build artifacts or other files we want to skip over? Not sure if you want to handle that here, elsewhere, or at all.

jonlouie (changes requested):
Minor questions and comments, otherwise looks good

src/CTA.WebForms2Blazor/Services/WorkspaceManagerService.cs
comsemer: These two functions exist for use by FileInformation objects or ClassInformation objects so that they can pause their execution if a critical condition isn't met yet. For instance, we want to hold off on adding documents to the workspace and hold off on starting code migrations until the solution and projects have been added to both workspaces. Not sure if there's a better way to do it outside of Task.Delay in a loop that checks the condition though

src/CTA.WebForms2Blazor/ProjectManagement/ProjectAnalyzer.cs
comsemer: Yeah we should definitely handle that, probably here as well, but not sure how we want to go about it yet. For now I'm just probably just going to put a todo comment in the method so we can come back to it

src/CTA.WebForms2Blazor/Services/WorkspaceManagerService.cs
jonlouie: Got it, I understand now. I’m not sure of a better way at the moment but we can revisit later if we need to. Should be ok for now

src/CTA.WebForms2Blazor/ProjectManagement/ProjectAnalyzer.cs
jonlouie: Cool, sounds good. We can handle this after discussing potential options

jonlouie (approved):

src/CTA.WebForms2Blazor/Services/WorkspaceManagerService.cs
```diff
@@ -0,0 +1,153 @@
...
+        public Solution CurrentSolution { get { return _workspace.CurrentSolution; } }
+
+        public void CreateSolutionFile()
+        {
+            if (_workspace != null)
+            {
+                throw new InvalidOperationException(WORKSPACE_DUPLICATE_ERROR);
+            }
+
+            _workspace = new AdhocWorkspace();
```
KallanBao, line 37: What is an AdhocWorkspace?

KallanBao (commented):
Looks good to me, besides the waiting while loop but like you mentioned before I'm not sure what the better way to approach that is yet

KallanBao (approved):

src/CTA.WebForms2Blazor/Services/WorkspaceManagerService.cs
comsemer: It's basically a container for solutions, projects, and code documents that we can use to create syntax trees and semantic models. The other workspace type we could use is the MSBuildWorkspace which would be better for the WebForms project but its not available in netcore3.1 which cta uses. Adhoc is nice because it allows full modification of solutions and projecs though, which MSBuild doesn't. Here are the MS docs on Adhoc workspaces https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.adhocworkspace?view=roslyn-dotnet-3.9.0

Eruanion (commented):
Looks good, good work.

src/CTA.WebForms2Blazor/Services/WorkspaceManagerService.cs
Eruanion: We could end up with infinite loop here if somehow condition is never met, can we have an escape condition if this runs for too long?

PR #139, opened 2021-06-15
Implemented StaticFileConverter to migrate static files

KallanBao:
## Description
* Implemented StaticFileConverter class to migrate static files to wwwroot directory
* Code refactoring, FileInformation objects now FileConverters, created new FileInformation objects

## Supplemental testing
Basic unit testing to make sure file byte array representation is generated and relative path is changed accordingly

---


src/CTA.WebForms2Blazor/FileConverters/StaticFileConverter.cs
```diff
@@ -0,0 +1,37 @@
...
+            var workingDirectory = Environment.CurrentDirectory;
+            var testProjectPath = Directory.GetParent(workingDirectory).Parent.Parent.FullName;
+            var path = Path.Combine(testProjectPath, RelativePath);
+            return File.ReadAllBytes(path);
+        }
+
+        public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
+        {
+            var newPath = Path.Combine("wwwroot", RelativePath);
+            FileInformation fi = new StaticFileInformation(newPath, GetFileBytes());
```
comsemer, line 29: This can just create an instance of FileInformation once it is no longer abstract, there's no need to use a subclass if it offers no extra functionality

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using CTA.WebForms2Blazor.FileConverters;
+
+namespace CTA.WebForms2Blazor.Tests
+{
+    [TestFixture]
+    class FileConverterTests
+    {
+        private const string TestFilesDirectoryPath = "TestingArea\\TestFiles";
```
jonlouie, line 18: Same as above, use `Path.Combine()`

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -0,0 +1,57 @@
...
+            _testStaticFilePath = Path.Combine(TestFilesDirectoryPath, "SampleStaticFile.png");
+            _testViewFilePath = Path.Combine(TestFilesDirectoryPath, "SampleViewFile.aspx");
+            _testProjectFilePath = Path.Combine(TestFilesDirectoryPath, "SampleProjectFile.csproj");
+        }
+
+        [SetUp]
+        public void Setup()
+        {
+
+        }
```
jonlouie, line 42: Can remove if not being used. If being saved for future use, leave a quick comment inside

comsemer (changes requested):
Mostly good, I just found some general things that I think should be altered regarding the new File Information and File Converter classes

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            FileConverter fc = new StaticFileConverter(_testStaticFilePath);
+
+            IEnumerable<FileInformation> fileList = await fc.MigrateFileAsync();
+            FileInformation fi = fileList.First();
+            byte[] bytes = fi.FileBytes;
+
+            Assert.IsTrue(bytes.Length != 0);
+            Assert.IsTrue(fi.RelativePath.Equals("wwwroot\\TestingArea\\TestFiles\\SampleStaticFile.png"));
+        }
```
jonlouie, line 55: A couple of suggestions for this test:

1. Rename test to `MigrateFileAsync_Moves_All_File_Bytes_To_wwwroot_Directory`. The ideal test name will include the function being tested, the condition of the test, and the expected outcome (e.g. `MyFunction_ThrowsException_With_Null_Parameter()`). Sometimes this is impossible or too wordy, so we can compromise based on context.
2. In the first assertion, we want to specify the exact number of bytes (e.g. what if the original file is 1 byte but the migrated file is 2 bytes?). Get the number of bytes of the original file and compare that to `fi.FileBytes.Length`.
3. Use `Path.Combine()` in the second `Assert` statement
4. Consider creating 1 test file per class, e.g. StaticFileConverterTests.cs. This will keep tests organized and prevent confusion if each converter class tests a shared interface with different implementations.

src/CTA.WebForms2Blazor/FileInformationModel/ConfigFileInformation.cs
```diff
@@ -1,25 +1,14 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
-using System.Threading.Tasks;
 
 namespace CTA.WebForms2Blazor.FileInformationModel
 {
-    public class ConfigFileInformation : FileInformation
+    class ConfigFileInformation : FileInformation
```
jonlouie, line 7: Should this class be public?

src/CTA.WebForms2Blazor/FileInformationModel/CodeFileInformation.cs
```diff
@@ -1,51 +1,14 @@
...
 using System.Collections.Generic;
-using System.IO;
-using System.Threading.Tasks;
-using CTA.WebForms2Blazor.Services;
+using System.Text;
 
 namespace CTA.WebForms2Blazor.FileInformationModel
 {
-    public class CodeFileInformation : FileInformation
+    class CodeFileInformation : FileInformation
```
jonlouie, line 7: Should this class be public?

src/CTA.WebForms2Blazor/FileConverters/FileConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+        public string RelativePath { get { return _relativePath; } }
+
+        protected FileConverter(string relativePath)
+        {
+            _relativePath = relativePath;
+        }
+
+        public abstract Task<IEnumerable<FileInformation>> MigrateFileAsync();
+
+        protected abstract byte[] GetFileBytes();
```
jonlouie, line 20: Have you and @comsemer discussed if getting the file bytes will differ based on the type of file being processed? My thought is that it would be identical for all files in which case `GetFileBytes()` can become a public method with an implementation. This would save work on each child class of `FileConverter`

src/CTA.WebForms2Blazor/Factories/FileConverterFactory.cs
```diff
@@ -36,29 +36,29 @@ public FileInformation Build(FileInfo document)
             string relativePath = Path.GetRelativePath(_sourceProjectPath, document.FullName);
             string extension = document.Extension;
 
-            FileInformation fi;
+            FileConverter fi;
```
jonlouie, line 39: Rename `fi` to `fc` to match previous naming convention or to `fileConverter` as another suggestion.

jonlouie (changes requested):

src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            _webFormsWorkspaceBuilder.NotifyNewExpectedDocument();
+
+            // TODO: Create class information objects and call
+            // _blazorProjectBuilder.NotifyNewExpectedDocument() in
+            // each one, this is because potentially multiple files
+            // are produced by this file information type
+        }
+
+        protected override byte[] GetFileBytes()
```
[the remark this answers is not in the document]
KallanBao: It won't let it be private since its inherited from an abstract class, the only way to make unaccessible from the other classes is making it protected

src/CTA.WebForms2Blazor/FileInformationModel/FileInformation.cs
```diff
@@ -1,21 +1,22 @@
-﻿using System.Threading.Tasks;
+﻿using System;
 using System.Collections.Generic;
+using System.Text;
 
 namespace CTA.WebForms2Blazor.FileInformationModel
 {
     public abstract class FileInformation
```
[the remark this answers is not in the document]
KallanBao: I mean different file information models can store different information right? Like config model might need to have separate information on which files its configurations are for

src/CTA.WebForms2Blazor/FileConverters/FileConverter.cs
KallanBao: I thought some of the other types of files we might do some processing before converting to byte[], but I can discuss again to confirm

src/CTA.WebForms2Blazor/FileInformationModel/FileInformation.cs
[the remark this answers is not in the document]
comsemer: Well the information they store is different, yes, but all of the information for the file is in bytes[], so the format of the information is the same. The point of the file information objects are to have a unified format to represent all files that the project builder can use indiscriminately

src/CTA.WebForms2Blazor/FileConverters/FileConverter.cs
Eruanion: We can also provide a base implementation and only override it if needed for exceptions. Another approach would be to store a base version of the implementation and reference that in all common method and use different code for the exceptions.

Eruanion (changes requested):
Good work, just need to review other people's suggestions and comments.

src/CTA.WebForms2Blazor/MigrationManager.cs
```diff
@@ -21,7 +21,7 @@ public class MigrationManager
 
         private WorkspaceManagerService _webFormsWorkspaceManager;
         private WorkspaceManagerService _blazorWorkspaceManager;
-        private FileInformationFactory _fileFactory;
+        private FileConverterFactory _fileFactory;
```
comsemer, line 24: I think Jon made a comment about changing the name to _converterFactory on the last version

comsemer (changes requested):
Looks great, just that one field left that needs to be renamed

comsemer (approved):
Perfect

src/CTA.WebForms2Blazor/FileConverters/StaticFileConverter.cs
```diff
@@ -15,18 +15,14 @@ public StaticFileConverter(string relativePath) : base(relativePath)
...
-        protected override byte[] GetFileBytes()
+        public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
         {
+            var newPath = Path.Combine("wwwroot", RelativePath);
             var workingDirectory = Environment.CurrentDirectory;
             var testProjectPath = Directory.GetParent(workingDirectory).Parent.Parent.FullName;
-            var path = Path.Combine(testProjectPath, RelativePath);
-            return File.ReadAllBytes(path);
-        }
+            var fullPath = Path.Combine(testProjectPath, RelativePath);
```
jonlouie: Does `testProjectPath` represent something from a test environment? If so, can we update this so our logic does not rely on a test environment?

jonlouie (commented):
In addition to what @comsemer mentioned, I added one more question that I'm unsure of.

src/CTA.WebForms2Blazor/FileConverters/StaticFileConverter.cs
KallanBao: Oh it doesn't rely on some test environment, it should just be the current directory. Its only named that way since I copy pasted the logic from somewhere else but I can fix that

src/CTA.WebForms2Blazor/FileConverters/StaticFileConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+        public StaticFileConverter(string relativePath) : base(relativePath)
+        {
+
+        }
+
+        public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
+        {
+            var newPath = Path.Combine("wwwroot", RelativePath);
+            var workingDirectory = Environment.CurrentDirectory;
+            var testProjectPath = Directory.GetParent(workingDirectory).Parent.Parent.FullName;
```
Eruanion, line 22: Also concerned about jon's comment here but this appears to be the context prefix for the path rather than actual test project path, perhaps just a naming oversight?

tst/CTA.WebForms2Blazor.Tests/Factories/FileConverterFactoryTests.cs
```diff
@@ -5,13 +5,14 @@
...
+using CTA.WebForms2Blazor.FileConverters;
 
 namespace CTA.WebForms2Blazor.Tests.Factories
 {
     [TestFixture]
-    class FileInformationFactoryTests
+    class FileConverterFactoryTests
     {
-        private const string TestFilesDirectoryPath = "TestingArea/TestFiles";
+        private const string TestFilesDirectoryPath = "TestingArea\\TestFiles";
```
Eruanion, line 15: can we use path.combine as Jon suggested before so we can support multiple platforms?

Eruanion (changes requested):
Just a couple of things left.

src/CTA.WebForms2Blazor/FileConverters/StaticFileConverter.cs
KallanBao: Yep, I'll rename it!

src/CTA.WebForms2Blazor/FileConverters/StaticFileConverter.cs
jonlouie: Got it, no problem. I think renaming it to reduce confusion should be sufficient then.

tst/CTA.WebForms2Blazor.Tests/Factories/FileConverterFactoryTests.cs
KallanBao: Whoops, missed this one

Eruanion (approved):

jonlouie (approved):

PR #141, opened 2021-06-16
Roslyn Extensions and Class Converters, Factory

comsemer:
## Description
* Document-based syntax tree retrieval
* Document-based semantic model retrieval
* Syntax tree extension methods to retrieve namespace level type references
* Semantic model extension methods to retrieve all referenced namespaces for a given type and retrieve all inherited base types of a given type
* Class Converter and Subclasses
* Class Converter factory and build implementation

---


tst/CTA.WebForms2Blazor.Tests/Extensions/SyntaxTreeExtensionTests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            var syntaxTree = await _workspaceManager.GetCurrentDocumentSyntaxTree(did1);
+            var testTypeNodes = syntaxTree.GetRoot().DescendantNodes().OfType<TypeDeclarationSyntax>();
+            var testType1 = testTypeNodes.Single(node => node.Identifier.ToString().Equals("TestType1"));
+            var testType2 = testTypeNodes.Single(node => node.Identifier.ToString().Equals("TestType2"));
+
+            var types = syntaxTree.GetNamespaceLevelTypes();
+            
+            Assert.True(types.Contains(testType1));
+            Assert.True(types.Contains(testType2));
```
jonlouie, line 59: More of a heads up than a suggestion, but for collections you can use something like:

```csharp
var expectedTypes = new List<TypeDeclarationSyntax> 
{ 
    testType1,
    testType2
};
CollectionAssert.AreEquivalent(expectedTypes, types);
```

tst/CTA.WebForms2Blazor.Tests/Factories/ClassConverterFactoryTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        [Test]
+        public async Task BuildMany_Splits_Classes()
+        {
+            var testDocumentPath = Path.Combine("C:", "Directory1", "Directory2", "TestDocumentName.cs");
+            var did = _workspaceManager.AddDocument(_primaryProjectId, "TestDocument", DocumentMultiClassText);
+            var model = await _workspaceManager.GetCurrentDocumentSemanticModel(did);
+            var classConverters = _classConverterFactory.BuildMany(testDocumentPath, model);
+
+            Assert.AreEqual(classConverters.Count(), 4);
```
jonlouie, line 112: nit: Switch the parameter values in `AreEqual()`.

The signature of the assertion is `AreEqual(object expected, object actual)` which is important because if the assertion fails, it will print out:
```
Expected value: {expected}
Actual value: {actual}
```

So having the values switched could cause confusion.

tst/CTA.WebForms2Blazor.Tests/Services/WorkspaceManagerServiceTests.cs
```diff
@@ -188,15 +198,94 @@ public void WaitUntilAllDocumentsInWorkspace_Task_Reflects_Cancellations()
...
+                                            SyntaxFactory.Identifier("TestMethod1"))
+                                        .WithModifiers(
+                                            SyntaxFactory.TokenList(
+                                                SyntaxFactory.Token(SyntaxKind.PublicKeyword)))
+                                        .WithBody(
+                                            SyntaxFactory.Block())))))))
+                .NormalizeWhitespace().SyntaxTree;
+
+            Assert.True(syntaxTree.IsEquivalentTo(targetSyntaxTree));
+        }
```
jonlouie, line 248: A simpler alternative to comparing syntax trees is to compare the string output. For example:
`Assert.AreEqual(TestDocument1Text, syntaxTree.ToFullString())`

In my opinion, the alternative is easier to read/understand. They are effectively equivalent so feel free to change or keep the same.

jonlouie (changes requested):

src/CTA.WebForms2Blazor/ClassConverters/ClassConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+            INamedTypeSymbol originalClassSymbol)
+        {
+            _relativePath = relativePath;
+            _sourceFileSemanticModel = sourceFileSemanticModel;
+            _originalClassSymbol = originalClassSymbol;
+        }
+
+        public abstract Task MigrateClassAsync();
+    }
+}
```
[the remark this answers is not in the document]
comsemer: These are for doing conversions of individual classes, the point of the class converters and factory are to split a single code file into all namespace level types then process them and generate a new file for each, in this case a CodeFileConverter would return the combined results of multiple ClassConverter MigrateClassAsync calls. Also, now that I'm looking at it, MigrateClassAsync should return a Task<FileInformation> not just Task

jonlouie (approved):

KallanBao (approved):

PR #143, opened 2021-06-17
Fix resource concurrency issues and other minor issues

### LINKED ISSUE #142: Fix reading template files concurrency issues
## Description of issue
When downloading and extracting files used as templates, we sometimes run into concurrency issues. We need to improve this process by downloading the files independently

marknfawaz:
## Related issue

Closes #142

### DIFF
--- a/src/CTA.Rules.Config/Constants.cs
+++ b/src/CTA.Rules.Config/Constants.cs
@@ -33,7 +33,8 @@ namespace CTA.Rules.Config
 
         public const string S3RootUrl = "https://s3.us-west-2.amazonaws.com/aws.portingassistant.dotnet.datastore";
         public const string S3RecommendationsBucketUrl = S3RootUrl + "/recommendationsync/recommendation";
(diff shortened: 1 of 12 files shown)


jonlouie (approved):

src/CTA.Rules.Config/Constants.cs
```diff
@@ -98,5 +99,18 @@ public class Constants
         public const string ConnectionStrings = "ConnectionStrings";
 
         public static readonly List<string> SupportedMethodModifiers = new List<string>() { "public", "internal", "protected", "private", "abstract", "extern", "override", "static", "unsafe", "virtual", "async" };
+
+
```
Eruanion, line 103: Do you think we should use enums for these file names? If we get more templates in the future I would prefer to keep typos from making their way in here and it would also be easier to locate this code by looking all references on the enums when we add new enum types.

Eruanion (commented):
Justr a small comment, looks good.

PR #146, opened 2021-06-21
Implemented part of ConfigFileConverter to Migrate web.config using CTA

KallanBao:
## Description
* Added CTA configMigrate method to Webforms2Blazor project to migrate web.config files to appsettings.json

## Supplemental testing
Tested migrating eShopOnWebforms web.config file to Blazor. Generated file not exactly the same but still works

---


src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
```diff
@@ -3,19 +3,57 @@
...
+        private string _projectPrefix;
+        private string _fullPath;
+        private string _relativeDirectory;
+
         public ConfigFileConverter(string relativePath) : base(relativePath)
         {
+            var workingDirectory = Environment.CurrentDirectory;
+            _projectPrefix = Directory.GetParent(workingDirectory).Parent.Parent.FullName;
+            _fullPath = Path.Combine(_projectPrefix, RelativePath);
+            _relativeDirectory = Path.GetDirectoryName(RelativePath);
```
comsemer, line 29: Environment.CurrentDirectory gives the working directory of the WebForms2Blazor project, not the WebFormsProject we're trying to migrate, if you need access to the full path maybe modifying the FileConverters to take both the relative path and the input base path as parameters is a good idea

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
```diff
@@ -3,19 +3,57 @@
...
+using System.IO;
+using System.Reflection;
+using Newtonsoft.Json;
+using System.Linq;
 
 namespace CTA.WebForms2Blazor.FileConverters
 {
     public class ConfigFileConverter : FileConverter
     {
+        private const string webConfigFile = "web.config";
```
comsemer, line 19: I believe const variables should be PascalCase

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
```diff
@@ -3,19 +3,57 @@
...
 
-        public override Task<IEnumerable<FileInformation>> MigrateFileAsync()
+
+        public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
         {
-            throw new NotImplementedException();
+            string filename = Path.GetFileName(RelativePath);
+            FileInformation fi;
+
+            if (filename.Equals(webConfigFile))
```
comsemer, line 39: You may want to make this comparison case insensitive, I had to do a similar thing when implementing the class converter factory class i.e. `filename.Equals(webConfigFile, StringComparison.InvariantCultureIgnoreCase)`

src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
```diff
@@ -22,6 +22,15 @@ public ConfigMigrate(string projectDir, ProjectType projectType)
             _projectType = projectType;
             _hasData = false;
         }
+
+        public void WebformsWebConfigMigrateHelper()
+        {
+            Configuration configXml = LoadWebConfig(_projectDir);
+            var templateContent = @"{}";
+            JObject config = ProcessWebConfig(configXml, templateContent);
+            AddAppSettingsJsonFile(config, _projectDir);
```
comsemer, line 31: If you make this return `config.ToString()` we can just use that content directly instead of calling `AddAppSettingsJsonFile` and having to re-read then write that content

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
```diff
@@ -3,19 +3,57 @@
...
+            {
+                //ProjectType WebForms doesn't really exist yet, but can be added for more specific configuration
+                ConfigMigrate configMigrate = new ConfigMigrate(_fullPath, ProjectType.WebForms);
+                configMigrate.WebformsWebConfigMigrateHelper();
+
+                string newPath = Path.Combine(_relativeDirectory, "appsettings.json");
+                string fullNewPath = Path.Combine(_projectPrefix, newPath);
+                fi = new FileInformation(newPath, File.ReadAllBytes(fullNewPath));
```
comsemer, line 47: See first comment about change to no longer have to re-read then write content like this

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
```diff
@@ -3,19 +3,57 @@
...
+        public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
         {
-            throw new NotImplementedException();
+            string filename = Path.GetFileName(RelativePath);
+            FileInformation fi;
+
+            if (filename.Equals(webConfigFile))
+            {
+                //ProjectType WebForms doesn't really exist yet, but can be added for more specific configuration
+                ConfigMigrate configMigrate = new ConfigMigrate(_fullPath, ProjectType.WebForms);
```
comsemer, line 42: This only handles 1 config file but WebForms projects can have multiple Web.Config files, are you planning to implement multi-file support later?

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
comsemer: Also, I believe there can be config files other than web.config and package.config so we do we need to add some kind of processing for those too?

comsemer (changes requested):

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
KallanBao: True, I'll change that

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
KallanBao: Well the name should still be web.config, but they may be in different modules. This should convert each web.config file seperately but it won't merge them together. I'm not sure what the expected behavior is tbh, are multiple web.config files supposed to end up all in one appsettings.json file are should there be multiple appsettings.json files as well

src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
comsemer: Well you don't have to do the byte conversion in there but you can if you want. I just meant that we should directly return the conversion result instead of writing it to disk then re-reading and re-writing

src/CTA.WebForms2Blazor/FileConverters/ConfigFileConverter.cs
comsemer: I'm pretty sure there should only be 1 appsettings.json file

jonlouie: Staying in line with my comment above, let's aim to process just a single web.config, at least to start.

tst/CTA.Rules.Test/Actions/ActionHelpers/ConfigMigrateTests.cs
```diff
@@ -11,6 +11,84 @@ namespace CTA.Rules.Test.Actions.ActionHelpers
...
+
+            // Invoke method and read contents of method output
+            var content = (JObject)processWebConfigMethod.Invoke(configMigrateInstance, new object[] { configuration, templateContent });
+
+            var methodParams = new object[] { content, outputDir };
+            addAppSettingsJsonFileMethod.Invoke(configMigrateInstance, methodParams);
+
+            var appSettingsContent = File.ReadAllText(Path.Combine(outputDir, "appsettings.json"));
+            File.Delete(Path.Combine(outputDir, "appsettings.json"));
+            File.Delete(Path.Combine(outputDir, "web.config"));
```
jonlouie, line 86: If modifying the file system in unit tests, use a `[OneTimeSetUp]` function to create a directory containing all test files and write files there. Then create a `[OneTimeTearDown]` function to delete the directory you created. This is a safe way to ensure that test files are always cleaned up (OneTimeTearDown is always called at the end of tests, even if an error occurs during a test). 

Note: The test class will need to be decorated with a `[TestFixture]` attribute as well in order for `[OneTime*]` attributes to take effect. I believe there are other examples of this in the project available for reference as well.

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -20,21 +20,23 @@ class FileConverterTests
...
             _testCodeFilePath = Path.Combine(TestFilesDirectoryPath, "TestClassFile.cs");
-            _testConfigFilePath = Path.Combine(TestFilesDirectoryPath, "SampleConfigFile.config");
+            _testWebConfigFilePath = Path.Combine(TestFilesDirectoryPath, "web.config");
             _testStaticFilePath = Path.Combine(TestFilesDirectoryPath, "SampleStaticFile.png");
             _testViewFilePath = Path.Combine(TestFilesDirectoryPath, "SampleViewFile.aspx");
             _testProjectFilePath = Path.Combine(TestFilesDirectoryPath, "SampleProjectFile.csproj");
+            _testAreaFullPath = Path.Combine(_testProjectPath, TestFilesDirectoryPath);
         }
 
         [SetUp]
```
jonlouie, line 42: Delete this `[SetUp]` function if not being used

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -57,5 +59,27 @@ public async Task TestStaticFileConverter()
...
+            string fullTestPath = Path.Combine(_testAreaFullPath, "Test_appsettings.json");
+
+            var appSettingsContent = File.ReadAllText(resPath);
+
+            Assert.True(appSettingsContent.Contains("UseMockData"));
+            Assert.True(appSettingsContent.Contains("UseCustomizationData"));
+            Assert.True(appSettingsContent.Contains("CatalogDBContext"));
+
+            //Not valid since different appsettings.json files can both work
+            //Assert.IsTrue(bytes.Length == new FileInfo(fullTestPath).Length);
```
jonlouie, line 81: Can delete if not being used

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -57,5 +59,27 @@ public async Task TestStaticFileConverter()
...
             Assert.IsTrue(fi.RelativePath.Equals(Path.Combine("wwwroot", _testStaticFilePath)));
         }
+
+        [Test]
+        public async Task TestWebConfigFileConverter()
+        {
+            FileConverter fc = new ConfigFileConverter(_testWebConfigFilePath);
+            IEnumerable<FileInformation> fileList = await fc.MigrateFileAsync();
+            FileInformation fi = fileList.First();
+            byte[] bytes = fi.FileBytes;
```
jonlouie, line 69: Can delete `byte[] bytes = fi.FileBytes;` if not being used

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -57,5 +59,27 @@ public async Task TestStaticFileConverter()
...
+
+            var appSettingsContent = File.ReadAllText(resPath);
+
+            Assert.True(appSettingsContent.Contains("UseMockData"));
+            Assert.True(appSettingsContent.Contains("UseCustomizationData"));
+            Assert.True(appSettingsContent.Contains("CatalogDBContext"));
+
+            //Not valid since different appsettings.json files can both work
+            //Assert.IsTrue(bytes.Length == new FileInfo(fullTestPath).Length);
+            File.Delete(resPath);
```
jonlouie, line 82: Similar as above, need to find a safe way to delete this in case test fails. Maybe wrap entire test in a `try` block and delete the expected file in the `finally` block (if the file exists).

jonlouie (changes requested):

tst/CTA.Rules.Test/Actions/ActionHelpers/ConfigMigrateTests.cs
KallanBao: I was following the structure of the rest of the tests in the file, that test was only added for me to see how ConfigMigrate worked. I can remove it if necessary though

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
KallanBao: Added a comment that it might be used later for future file converter tests

src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
```diff
@@ -22,6 +22,15 @@ public ConfigMigrate(string projectDir, ProjectType projectType)
             _projectType = projectType;
             _hasData = false;
         }
+
+        public string WebformsWebConfigMigrateHelper()
+        {
+            Configuration configXml = LoadWebConfig(_projectDir);
+            var templateContent = @"{}";
```
jonlouie, line 29: Let's replace templateContent with:

```csharp
var templateContent = TemplateHelper.GetTemplateFileContent(string.Empty, ProjectType.WebClassLibrary, Constants.AppSettingsJson);
```

What this will do is load an appsettings.json template file into templateContent- this will keep us in line with the existing behavior for other project types. This template is downloaded from our S3 templates bucket when a `SolutionPort` object is instantiated.

src/CTA.WebForms2Blazor/Factories/FileConverterFactory.cs
```diff
@@ -13,6 +14,13 @@ public class FileConverterFactory
         private readonly WorkspaceManagerService _blazorWorkspaceManager;
         private readonly WorkspaceManagerService _webFormsWorkspaceManager;
         private readonly ClassConverterFactory _classConverterFactory;
+        
+        public readonly HashSet<string> StaticResourceExtensions = new HashSet<string>
```
comsemer, line 18: Its possible that extensions outside of this list ought to be placed in the wwwroot folder. For instance, the customer may have downloadable .csv, .pdf, or .docx files. As far as I know there isn't a surefire way to determine which files ought to be placed in the folder and which ones shouldn't, not sure if guessing like this or just tossing all files in wwwroot is the right approach though

tst/CTA.Rules.Test/Actions/ActionHelpers/ConfigMigrateTests.cs
jonlouie: If this test is not confirming correctness, then we can delete it. If it is, then it should be ok to leave in

src/CTA.WebForms2Blazor/Factories/FileConverterFactory.cs
KallanBao: Yeah this isn't the best way to do it, but at least its good enough for demoing for eShopOnBlazor. In the future we might need to add a note for the customer to make sure and check if the correct files are in wwwroot

comsemer: True, or using some kind of optional blacklist configured by the customer before the run could be used, but this works for now

comsemer (approved):
Looks good

src/CTA.WebForms2Blazor/Factories/FileConverterFactory.cs
Eruanion: I think we are ok with this for now. For scope of project.

tst/CTA.Rules.Test/Actions/ActionHelpers/ConfigMigrateTests.cs
Eruanion: You can keep it if it adds code coverage for tests, if it was purely for you to debug just keep it locally without checking it in on the branch for your own benefit. If you think however it does provide  useful unit test coverage please keep it but implement it as Jon mentioned with a OnetimeTearDown attribute to ensure these always get cleaned up. \cta\tst\CTA.Rules.Test\SetupTests.cs has an example of this if you want to use it.

src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
KallanBao: I'm having an error when switching templateContent, saying there is no appsettings.json template to load. Do I have to create the template myself?

jonlouie (approved):

src/CTA.Rules.Actions/ActionHelpers/ConfigMigrate.cs
jonlouie: Per our conversation, the issue was a pathing error which we were able to resolve.

PR #151, opened 2021-06-23
Startup and Middleware syntax-tree-based template building

comsemer:
## Description
* Startup and Middleware syntax-tree-based template building
* Extra syntax tree extension methods
* Other useful syntax tree helper methods for dealing with other code files
---


jonlouie (changes requested):

tst/CTA.WebForms2Blazor.Tests/Helpers/RuntimeInjectableTests.cs
```diff
@@ -0,0 +1,89 @@
+﻿using CTA.WebForms2Blazor.Helpers;
+using NUnit.Framework;
+using Microsoft.CodeAnalysis;
+
+namespace CTA.WebForms2Blazor.Tests.Helpers
+{
+    [TestFixture]
+    public class RuntimeInjectableTests
+    {
+        private const string TestInjectableNamePascalCase = "TestName";
```
Eruanion, line 10: Is it necessary to test the camel case and the pascal case? These are usually syntax preferences for development rather than actual functional tests no? We can keep them there I was just wondering if there was any other thoughts behind these tests.

Eruanion (commented):
Good code, looks great.

tst/CTA.WebForms2Blazor.Tests/Helpers/RuntimeInjectableTests.cs
comsemer: The reason I chose to include these tests is mainly to make sure that when we reference some form of these runtime injectables we use the correct names. Also following these naming conventions likely benefits the customer

jonlouie (approved):
LGTM

Eruanion (approved):

KallanBao (approved):

PR #152, opened 2021-06-24
Implemented Basic Project File Migration

KallanBao:
## Description
* Implemented ProjectFileConverter such that basic Webforms csproj framework is created and PackageReferences are added. Still need to add references to the static files in the projectFile, but this is a start. Also, Project Sdk type and PackageReferences are hardcoded for now, will need to pass in actual values from CTA later, but this will work for demo.

---


src/CTA.Rules.ProjectFile/ProjectFileCreator.cs
```diff
@@ -133,16 +130,31 @@ public bool Create()
                 }
                 else
```
comsemer, line 131: I'm concerned about the if block right before this, if `IsCoreProject()` returns true then doesn't that mean that `CreateContents()` will return null? I'm not sure it matters for w2b but will this break normal CTA calls to `Create()`? I only ask because it seems like normally `Create()` would only return false in the event of an exception being thrown whereas that wouldn't hold in this case

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -30,6 +36,73 @@ class FileConverterTests
...
+       
+            [OneTimeTearDown]
+            public void Cleanup()
+            {
+                DeleteDir(0);
+            }
+       
+            private void DeleteDir(int retries)
+            {
+                if (retries <= 10)
```
comsemer, line 90: Were you running into problems deleting the directories?

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -30,6 +36,73 @@ class FileConverterTests
...
+            {
+                Setup(GetType());
+                TempDir = GetTstPath(Path.Combine(new [] { "Projects", "Temp", "W2B" }));
+                DownloadTestProjects();
+       
+                EShopOnBlazorSolutionFilePath = CopySolutionFolderToTemp("eShopOnBlazor.sln", TempDir);
+                EShopOnBlazorSolutionPath = Directory.GetParent(EShopOnBlazorSolutionFilePath).FullName;
+
+                CopyFolder = Directory.GetParent(EShopOnBlazorSolutionPath).FullName;
+                //var tempHelp = Directory.EnumerateFiles(EShopOnBlazorProjectPath, "*.csproj", SearchOption.AllDirectories);
```
jonlouie, line 63: remove unused code

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -30,6 +36,73 @@ class FileConverterTests
...
+                if (retries <= 10)
+                {
+                    try
+                    {
+                        Directory.Delete(TempDir, true);
+                        Directory.Delete(CopyFolder, true);
+                    }
+                    catch (Exception)
+                    {
+                        Thread.Sleep(60000);
```
comsemer, line 99: 60 Seconds is a long time, is there a reason the wait is this high?

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -30,6 +36,73 @@ class FileConverterTests
         private string _testProjectFilePath;
         private string _testAreaFullPath;
 
+        private ProjectAnalyzer _webFormsProjectAnalyzer;
+        private WorkspaceManagerService _blazorWorkspaceManager;
+        
+        [SetUpFixture]
+        public class WebForms2BlazorTestFixture : AwsRulesBaseTest
```
comsemer, line 43: I would try to give this setup fixture a more descriptive name

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -111,5 +191,61 @@ public async Task TestWebConfigFileConverter()
...
+
+        [Test]
+        public async Task TestProjectFileConverter()
+        {
+            FileConverter fc = new ProjectFileConverter(WebForms2BlazorTestFixture.EShopOnBlazorSolutionPath,
+                WebForms2BlazorTestFixture.EShopLegacyWebFormsProjectPath,
+                _blazorWorkspaceManager, _webFormsProjectAnalyzer);
+
+            IEnumerable<FileInformation> fileList = await fc.MigrateFileAsync();
+            FileInformation fi = fileList.First();
```
comsemer, line 237: If you're only expecting one file, you might want to use `Single()` instead of `First()` that way it will throw an error if more than one file information object is present, but not entirely necessary to make this change

comsemer (changes requested):
Looks pretty good to me, most of my comments are just questions with 1 or 2 minor nitpicks

src/CTA.WebForms2Blazor/FileConverters/ProjectFileConverter.cs
```diff
@@ -1,33 +1,108 @@
...
+
+            // Now we can finally create the ProjectFileCreator and use it
+            var projectReferences = analyzerResult?.ProjectBuildResult?.ExternalReferences?.ProjectReferences.Select(p => p.AssemblyLocation).ToList();
+            var metaReferences = analyzerResult?.ProjectBuildResult?.Project?.MetadataReferences?.Select(m => m.Display).ToList();
+            var projectFileCreator = new ProjectFileCreator(FullPath, projectConfiguration.TargetVersions, packages, 
+                projectReferences, ProjectType.WebForms, metaReferences);
 
-        public override Task<IEnumerable<FileInformation>> MigrateFileAsync()
+            return projectFileCreator.CreateContents();
+        }
```
jonlouie, line 102: Per our discussion, refactor out most of this since we can assume CTA will pass in most of the information we need, then move the hard coded values to a Program.cs and/or unit tests to test this class

jonlouie (changes requested):
In addition to the other comments, please remove the `.DS_Store` files and add that file to the .gitgnore file

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
jonlouie: This is copied over from our main repository. It's a common pattern to retry deletion at the end of unit tests, but it's mostly there as a safety net.

comsemer: Ah ok, makes sense

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
jonlouie: That was a name I defaulted to. I agree, it can use a more descriptive name such as "DownloadTestProjectsFixture" or something. 

Also, this fixture and related fields should go in its own file.

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
Eruanion: Yep this should be more around 1 second or lower with a retry or 20 times or so should do the trick.

tst/CTA.WebForms2Blazor.Tests/Factories/FileConverterFactoryTests.cs
```diff
@@ -39,13 +40,13 @@ public void OneTimeSetup()
         [SetUp]
         public void Setup()
         {
-            var webFormsWorkspaceManager = new WorkspaceManagerService();
+            var webFormsProjectAnalyzer = new ProjectAnalyzer(_testProjectPath, null);
             var blazorWorkspaceManager = new WorkspaceManagerService();
 
             blazorWorkspaceManager.CreateSolutionFile();
-            webFormsWorkspaceManager.CreateSolutionFile();
+            //webFormsWorkspaceManager.CreateSolutionFile();
```
Eruanion, line 47: remove unused code also, repo has history to keep track of everything if ever needed.

src/CTA.Rules.ProjectFile/ProjectFileCreator.cs
KallanBao: Hmm, I'm not to sure how to handle this, will prob discuss with you guys tomorrow

KallanBao:
I also added a Program.cs file that can be used for demoing purposes. It's mostly done, just needs some more testing and tweaking to make sure it works.

tst/CTA.WebForms2Blazor.Tests/FileConverterTests.cs
```diff
@@ -38,74 +38,15 @@ class FileConverterTests
...
-                }
-            }
-        }
 
         [OneTimeSetUp]
         public void OneTimeSetup()
         {
+            WebForms2BlazorTestFixture = new DownloadTestProjectsFixture();
+            WebForms2BlazorTestFixture.Setup();
+
```
jonlouie: Fixtures shouldn't be instantiated and their setup and teardown methods don't need to be explicitly called since they are automatically executed whenever unit tests are run. If there's any confusion on how these OneTimeSetup/TearDown functions are called, I'd checkout the NUnit docs on TextFixtures and SetupFixtures.

Since the properties we need from the fixture are static, we can reference them statically instead of creating a fixture object.

jonlouie (changes requested):

tst/CTA.WebForms2Blazor.Tests/DownloadTestProjectsFixture.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    [SetUpFixture]
+    public class DownloadTestProjectsFixture : AwsRulesBaseTest
+    {
+        public string TempDir;
+        public string CopyFolder;
+        public string DownloadLocation;
+        public string EShopOnBlazorSolutionPath;
+        public string EShopOnBlazorSolutionFilePath;
+        public string EShopLegacyWebFormsProjectPath;
```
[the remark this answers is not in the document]
KallanBao: If I make it static then in my tests it says I can't access static variables in a nonstatic context

jonlouie:
Also, there are a couple of file conflicts that need to be resolved. You'll need to pull down the latest from `w2b-intern-dev`, resolve conflicts, and commit the merge.

jonlouie (approved):

Eruanion (approved):

comsemer (approved):

PR #153, opened 2021-06-25
Readied for demo and completed basic class converter functionality

comsemer:
## Description
* Added class converter basic functionality
* Readied for demo
---


src/CTA.WebForms2Blazor/ClassConverters/ControlCodeBehindClassConverter.cs
```diff
@@ -1,24 +1,52 @@
...
 
         public override async Task<FileInformation> MigrateClassAsync()
         {
-            throw new NotImplementedException();
+            // NOTE: For now we make no code modifications, just to be
+            // ready for the demo and produces files
+            var requiredNamespaces = _sourceFileSemanticModel.GetNamespacesReferencedByType(_originalDeclarationSyntax);
+            var usingStatements = CodeSyntaxHelper.BuildUsingStatements(requiredNamespaces.Select(namespaceSymbol => namespaceSymbol.Name));
+            var namespaceNode = CodeSyntaxHelper.BuildNamespace(_originalClassSymbol.ContainingNamespace.Name, _originalDeclarationSyntax);
+            var fileText = CodeSyntaxHelper.GetFileSyntaxAsString(namespaceNode, usingStatements);
```
jonlouie, line 33: Seems like these 4 lines are repeated in a few of the converters. Would it make sense to define a function in the parent class that returns this information in a new object called `SourecFileComponents` or something similar?

src/CTA.WebForms2Blazor/ClassConverters/ControlCodeBehindClassConverter.cs
```diff
@@ -1,24 +1,52 @@
...
+            // TODO: Potentially remove certain folders from beginning of relative path
+            var relativePathFolder = Path.GetDirectoryName(_relativePath);
+            var oldFileName = Path.GetFileName(_relativePath);
+            // Path.GetFileNameWithoutExtension only removes the .cs of .ascx.cs
+            // and so we have to remove the extension manually
+            var baseFileNameLength = oldFileName.Length - Constants.ControlCodeBehindExtension.Length;
+            // The code behinds are responsible for generating the new razor
+            // file and so we replace the extension with .razor
+            var newFileName = oldFileName.Substring(0, baseFileNameLength) + Constants.RazorFileExtension;
+            return Path.Combine(Constants.RazorComponentDirectoryName, relativePathFolder, newFileName);
```
jonlouie, line 49: A couple of things:

1. Since this logic is used in multiple places, consider creating a private method (or static utility method) to replace the extension.
2. Not sure if you're aware, but similar to WebForms, Blazor can also have a code-behind files (i.e. `.razor` for front-end and `.razor.cs` for code-behind). Not sure if this is something you've considered or if it fits into your design, but here is a [brief example](https://www.claudiobernasconi.ch/2020/02/08/blazor-components-how-to-split-template-and-csharp-code/) if you're unfamiliar.

src/CTA.WebForms2Blazor/Factories/FileConverterFactory.cs
```diff
@@ -20,7 +18,7 @@ public class FileConverterFactory
         {
```
Eruanion, line 18: We should eventually do as Jon suggested and replace these with categories such as markup files, coding files, images etc.. or we can also replace these with constants within those categories so we can have a word for each of these such as Constants.CSharpCodeFileExtension for .cs for example. This doe snot need to be done now we can just add it in  quip for todos for now.

Eruanion (commented):
I agree with Jon's comments, otherwise looks good.

src/CTA.WebForms2Blazor/Helpers/FilePathHelper.cs
```diff
@@ -0,0 +1,34 @@
...
+        public static string AlterFileName(string oldFilePath, string newFileName = null, string oldExtension = null, string newExtension = null)
+        {
+            var actualOldExtension = oldExtension ?? Path.GetExtension(oldFilePath);
+            var oldFileDirectory = Path.GetDirectoryName(oldFilePath);
+            var oldFileName = Path.GetFileName(oldFilePath);
+
+            if (!oldFileName.Contains(actualOldExtension))
+            {
+                throw new ArgumentException(string.Format(FileNameDoesNotContainExtensionError, oldFileName, actualOldExtension));
+            }
```
jonlouie, line 21: I suggest using `.EndsWith()` instead of `.Contains()` because there could be an edge case where `oldExtension` exists elsewhere in `oldFilePath`.

jonlouie (changes requested):
One minor comment to address, otherwise looks good

KallanBao (approved):

jonlouie (approved):

PR #155, opened 2021-06-30
Change NodeToken.Clone from shallow clone to deep clone

ShanshanHe:
## Related issue

Closes: #ISSUE_NUMBER_HERE

## Description
    With shallow clone, the cloned NodeToken object maintains a reference to
    the original NodeToken object's mutable variables. Any changes made to the
    original object's variables are reflected to the cloned ones. Eg.
    TextChanges variable of type List<TextChange> is a member of NodeToken
    class, with shallow clone, we observed the
    `ProjectResult.ProjectActions.FileActions` returned by
    `SolutionPort.Run()` has their `FileAction.NodeToken` field points to
    the same list of `TextChange` resulting duplicated results. Changing the
    clone to deep clone solves the issue.

## Supplemental testing
   Added test for to validate NodeToken deep clone.

---

### DIFF
--- a/src/CTA.Rules.Models/Actions/Attributeaction.cs
+++ b/src/CTA.Rules.Models/Actions/Attributeaction.cs
@@ -8,7 +8,7 @@ namespace CTA.Rules.Models
     {
         public Func<SyntaxGenerator, AttributeSyntax, AttributeSyntax> AttributeActionFunc { get; set; }
         public Func<SyntaxGenerator, AttributeListSyntax, AttributeListSyntax> AttributeListActionFunc { get; set; }
-
+        public new AttributeAction Clone() => (AttributeAction)this.MemberwiseClone();
         public override bool Equals(object obj)
         {
             var action = (AttributeAction)obj;
--- a/src/CTA.Rules.Models/Actions/Attributeaction.cs
+++ b/src/CTA.Rules.Models/Actions/Attributeaction.cs
@@ -8,7 +8,7 @@ namespace CTA.Rules.Models
     {
         public Func<SyntaxGenerator, AttributeSyntax, AttributeSyntax> AttributeActionFunc { get; set; }
         public Func<SyntaxGenerator, AttributeListSyntax, AttributeListSyntax> AttributeListActionFunc { get; set; }
-        public new AttributeAction Clone() => (AttributeAction)this.MemberwiseClone();
+
         public override bool Equals(object obj)
         {
             var action = (AttributeAction)obj;
--- a/src/CTA.Rules.Models/Tokens/NodeToken.cs
+++ b/src/CTA.Rules.Models/Tokens/NodeToken.cs
@@ -55,23 +55,23 @@ namespace CTA.Rules.Models.Tokens
         public NodeToken Clone()
         {
             NodeToken cloned = (NodeToken)this.MemberwiseClone();
-            cloned.TextChanges = cloned.TextChanges.Select(textChange => textChange.Clone()).ToList();
-            cloned.TargetCPU = cloned.TargetCPU.ToList();
-            cloned.AttributeActions = cloned.AttributeActions.Select(action => action.Clone()).ToList();
-            cloned.AttributeListActions = cloned.AttributeListActions.Select(action => action.Clone()).ToList();
-            cloned.ClassDeclarationActions = cloned.ClassDeclarationActions.Select(action => action.Clone()).ToList();
-            cloned.InterfaceDeclarationActions = cloned.InterfaceDeclarationActions.Select(action => action.Clone()).ToList();
-            cloned.MethodDeclarationActions = cloned.MethodDeclarationActions.Select(action => action.Clone()).ToList();
-            cloned.ElementAccessActions = cloned.ElementAccessActions.Select(action => action.Clone()).ToList();
-            cloned.MemberAccessActions = cloned.MemberAccessActions.Select(action => action.Clone()).ToList();
-            cloned.UsingActions = cloned.UsingActions.Select(action => action.Clone()).ToList();
-            cloned.IdentifierNameActions = cloned.IdentifierNameActions.Select(action => action.Clone()).ToList();
-            cloned.InvocationExpressionActions = cloned.InvocationExpressionActions.Select(action => action.Clone()).ToList();
-            cloned.NamespaceActions = cloned.NamespaceActions.Select(action => action.Clone()).ToList();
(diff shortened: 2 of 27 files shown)


src/CTA.Rules.Models/Actions/Attributeaction.cs
```diff
@@ -8,7 +8,7 @@ public class AttributeAction : GenericAction
     {
         public Func<SyntaxGenerator, AttributeSyntax, AttributeSyntax> AttributeActionFunc { get; set; }
         public Func<SyntaxGenerator, AttributeListSyntax, AttributeListSyntax> AttributeListActionFunc { get; set; }
-
+        public new AttributeAction Clone() => (AttributeAction)this.MemberwiseClone();
```
marknfawaz, line 11: Instead of having a Clone() in each action, can we add a generic clone in the base class? Something like:

public T Clone<T>() => (T)this.MemberwiseClone();

This way we can reuse it in all the child classes by passing the type we are cloning. This applies to all the *Action updates.

src/CTA.Rules.Models/Tokens/NodeToken.cs
```diff
@@ -52,7 +52,28 @@ public NodeToken()
...
+            cloned.ElementAccessActions = cloned.ElementAccessActions.Select(action => action.Clone()).ToList();
+            cloned.MemberAccessActions = cloned.MemberAccessActions.Select(action => action.Clone()).ToList();
+            cloned.UsingActions = cloned.UsingActions.Select(action => action.Clone()).ToList();
+            cloned.IdentifierNameActions = cloned.IdentifierNameActions.Select(action => action.Clone()).ToList();
+            cloned.InvocationExpressionActions = cloned.InvocationExpressionActions.Select(action => action.Clone()).ToList();
+            cloned.NamespaceActions = cloned.NamespaceActions.Select(action => action.Clone()).ToList();
+            cloned.ObjectCreationExpressionActions = cloned.ObjectCreationExpressionActions.Select(action => action.Clone()).ToList();
+            cloned.PackageActions = cloned.PackageActions.Select(action => action.Clone()).ToList();
+            cloned.ProjectLevelActions = cloned.ProjectLevelActions.Select(action => action.Clone()).ToList();
+            cloned.ProjectFileActions = cloned.ProjectFileActions.Select(action => action.Clone()).ToList();
```
marknfawaz, line 74: What would happen if any of these actions are empty or null? Any checks we need to add here?

marknfawaz (changes requested):

src/CTA.Rules.Models/Actions/Attributeaction.cs
ShanshanHe: Good point, will add the clone method to the base action class instead.

src/CTA.Rules.Models/Tokens/NodeToken.cs
ShanshanHe: That's good point. Looks like`NodeToken`'s constructor initializes all the actions to empty lists so they won't be null, unless there are places we are explicitly setting the action lists to null after `NodeToken` objects are initialized, I can add a null check to be safe. As for empty list, I think `Select` works with empty list and wont throw any exception, and it'll return an empty list, so should be safe there.

src/CTA.Rules.Models/Tokens/NodeToken.cs
```diff
@@ -52,7 +52,28 @@ public NodeToken()
...
         public List<ProjectLevelAction> ProjectFileActions { get; set; }
 
-        public NodeToken Clone() => (NodeToken)this.MemberwiseClone();
+        public NodeToken Clone()
+        {
+            NodeToken cloned = (NodeToken)this.MemberwiseClone();
+            cloned.TextChanges = cloned.TextChanges?.Select(textChange => textChange.Clone()).ToList();
+            cloned.TargetCPU = cloned.TargetCPU?.ToList();
+            cloned.AttributeActions = cloned.AttributeActions?.Select(action => action.Clone<AttributeAction>())?.ToList();
+            cloned.AttributeListActions = cloned.AttributeListActions?.Select(action => action.Clone<AttributeAction>())?.ToList();
```
marknfawaz, line 61: Should the second one be attributelist?

src/CTA.Rules.Models/Tokens/NodeToken.cs
```diff
@@ -52,7 +52,28 @@ public NodeToken()
...
+            cloned.MemberAccessActions = cloned.MemberAccessActions.Select(action => action.Clone<MemberAccessAction>()).ToList();
+            cloned.UsingActions = cloned.UsingActions.Select(action => action.Clone<UsingAction>()).ToList();
+            cloned.IdentifierNameActions = cloned.IdentifierNameActions.Select(action => action.Clone<IdentifierNameAction>()).ToList();
+            cloned.InvocationExpressionActions = cloned.InvocationExpressionActions.Select(action => action.Clone<InvocationExpressionAction>()).ToList();
+            cloned.NamespaceActions = cloned.NamespaceActions.Select(action => action.Clone<NamespaceAction>()).ToList();
+            cloned.ObjectCreationExpressionActions = cloned.ObjectCreationExpressionActions.Select(action => action.Clone<ObjectCreationExpressionAction>()).ToList();
+            cloned.PackageActions = cloned.PackageActions.Select(action => action.Clone()).ToList();
+            cloned.ProjectLevelActions = cloned.ProjectLevelActions.Select(action => action.Clone<ProjectLevelAction>()).ToList();
+            cloned.ProjectFileActions = cloned.ProjectFileActions.Select(action => action.Clone<ProjectLevelAction>()).ToList();
```
marknfawaz, line 74: Should this be projectfileactions?

marknfawaz (changes requested):

src/CTA.Rules.Models/Tokens/NodeToken.cs
ShanshanHe: `AttributeListActions`'s type is `List<AttributeAction>`, so `AttributeAction` should be correct here.

src/CTA.Rules.Models/Tokens/NodeToken.cs
ShanshanHe: Both `ProjectLevelActions` and `ProjectFileActions` are of same type: `List<ProjectLevelAction>` so `ProjectLevelAction` should be correct here.

marknfawaz (approved):

PR #156, opened 2021-07-01
Created framework for identifying and converting .aspx controls

KallanBao:
## Description
* Added functionality in ViewFileConverter class to identify nodes in .aspx files
* Implemented ControlConverter class framework to convert WebForms controls to Blazor
* Refactored FileConverter test category

## Supplemental testing
Testing with very preliminary implementation of asp:HyperLink control migration

---


src/CTA.WebForms2Blazor/ControlConverters/HyperLinkControlConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using HtmlAgilityPack;
+
+namespace CTA.WebForms2Blazor.ControlConverters
+{
+    public class HyperLinkControlConverter : ControlConverter
+    {
+        private Dictionary<String, String> _hyperLinkControlsMap = new Dictionary<string, string>() 
+            {["navigateurl"] = "href", ["cssclass"] = "class"};
+        private readonly string _name = "a";
```
comsemer, line 11: You may want to make this a const as it shouldn't be modified, even in the constructor

src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using HtmlAgilityPack;
+
+namespace CTA.WebForms2Blazor.ControlConverters
+{
+    public class ButtonControlConverter : ControlConverter
+    {
+        public ButtonControlConverter() : base()
```
comsemer, line 8: You don't need to make the call to base() if it doesn't do anything, if you just added it for use later though that's fine then

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -0,0 +1,18 @@
+﻿using HtmlAgilityPack;
+
+namespace CTA.WebForms2Blazor.ControlConverters
+{
+    public abstract class ControlConverter
+    {
+        protected ControlConverter()
+        {
+            //Constructor might not be needed
```
comsemer, line 9: It's probably true that we probably don't need the constructor, having it here in case we need it is probably fine but removing it wouldn't hurt either, we can just add it back later if needed

src/CTA.WebForms2Blazor/ControlConverters/HyperLinkControlConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            newAttributes = newAttributes.Trim();
+
+            string template = 
+                @"<{0} {1}>{2}</{0}>";
+            
+            string newContent = String.Format(template, _name, newAttributes, node.InnerHtml);
+            HtmlNode newNode = HtmlNode.CreateNode(newContent);
+            return newNode;
+        }
```
jonlouie, line 38: A couple of suggestions for this method: 
1. Can move this implementation to the ControlConverter class and replace references to `_hyperLinkControlsMap` and `_name` with their `abstract` replacements. This will allow us to reuse all of this code for each converter. If there's a converter that ends up requiring special behavior, we can always override this method.
2. Let's move everything in lines 20-30 into a separate method called `ConvertAttributes` or something similar. This method can also be defined in the ControlConverter class.
3. Lines 21-28 can also be refactored in a more concise manner using a LINQ function:
```csharp
var convertedAttributes = node.Attributes
    .Where(attr => AttributeMap.ContainsKey(attr.Name))
    .Select(attr => $"{AttributeMap[attr.Name]}={attr.Value}");
var convertedAttributesString = string.Join(" ", convertedAttributes);
```
4. Put lines 32-36 into a new method called `CreateBlazorNodeFromParts` or something similar. The parameters can be template, node name, convertedAttributes, and innerHtml and return type can be HtmlNode.

src/CTA.WebForms2Blazor/ControlConverters/HyperLinkControlConverter.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using HtmlAgilityPack;
+
+namespace CTA.WebForms2Blazor.ControlConverters
+{
+    public class HyperLinkControlConverter : ControlConverter
+    {
+        private Dictionary<String, String> _hyperLinkControlsMap = new Dictionary<string, string>() 
+            {["navigateurl"] = "href", ["cssclass"] = "class"};
```
comsemer, line 10: Something like css class seems like it would hold for all controls. If so then we might want to add this to some sort of secondary general control property map or have Convert2Blazor call the base class method where general changes like that are done

src/CTA.WebForms2Blazor/ControlConverters/HyperLinkControlConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+                {
+                    string newAttr = _hyperLinkControlsMap[attr.Name] + "=" + attr.Value + " ";
+                    newAttributes += newAttr;
+                }
+            }
+
+            newAttributes = newAttributes.Trim();
+
+            string template = 
+                @"<{0} {1}>{2}</{0}>";
```
comsemer, line 33: This template seems like it will be used in a lot of the control classes, may want to put it in a more central location

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -1,26 +1,93 @@
...
 using CTA.WebForms2Blazor.FileInformationModel;
+using CTA.WebForms2Blazor.Helpers;
+using HtmlAgilityPack;
 
 namespace CTA.WebForms2Blazor.FileConverters
 {
     public class ViewFileConverter : FileConverter
     {
-        public ViewFileConverter(string sourceProjectPath, string fullPath) : base(sourceProjectPath, fullPath)
+        private Dictionary<String, ControlConverter> ControlRulesMap; //From constructor or service
```
jonlouie, line 16: Can we make the ControlRulesMap a constant/static value? It can be placed in its own class or in a constants class. If we do this, we can remove it as a parameter from the constructor.

src/CTA.WebForms2Blazor/ControlConverters/HyperLinkControlConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public override HtmlNode Convert2Blazor(HtmlNode node)
+        {
+            string newAttributes = "";   
+            foreach (HtmlAttribute attr in node.Attributes)
+            {
+                if (_hyperLinkControlsMap.ContainsKey(attr.Name))
+                {
+                    string newAttr = _hyperLinkControlsMap[attr.Name] + "=" + attr.Value + " ";
```
comsemer, line 25: May want to make this into a template string like the one below, also this seems like it may be reused, maybe a function in the base class that takes in some number of control attribute maps and returns the formatted string would be good for this

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -1,26 +1,93 @@
...
+        //Tuple contains node to be converted, parent node, and instructions to convert node in that order
+        private List<(HtmlNode Node, HtmlNode Parent, ControlConverter Rules)> NodesToBeConverted;
```
jonlouie, line 18: Instead of using a `Tuple<HtmlNode, HtmlNode, ControlConverter>`, can we create a new class called `ControlConversionAction` or something similar? This could be a simple POCO class ([POCO example](https://stackoverflow.com/a/250006)). This will improve readability because at first glance, it's difficult to interpret what exactly the Tuple object means or how these objects will be used.

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -1,26 +1,93 @@
...
-            // View file converters don't need to return any file
-            // information objects, this is to be handled by the
-            // code-behinds for each
-            return Enumerable.Empty<FileInformation>();
+            // View file converters will return razor file contents with
+            // only view layer, code behind will be created in another file
+            HtmlDocument migratedDocument = GetRazorContents();
+            string contents = migratedDocument.DocumentNode.WriteTo();
+
+            string newFileName = "testing.razor"; //new file name depends, add later
```
comsemer, line 85: If you plan to change this later that's fine, but just {oldFileName}.razor should be fine. FilePathHelper has a method for this, but you will need to check what kind of view file this is too

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -1,26 +1,93 @@
...
+                return;
+            }
+
+            foreach (HtmlNode child in node.ChildNodes)
+            {
+                DfsPostorderWalker(child, node);
+            }
+
+            GetRules(node, parent);
+        }
```
jonlouie, line 56: Would improve readability to rename closer to what the method is doing, e.g. `FindConversionActions` or something similar. It would help to add a doc comment describing how the rules are being found (i.e. DFS through tree to identify nodes in postorder fashion).

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -1,26 +1,93 @@
...
-            // information objects, this is to be handled by the
-            // code-behinds for each
-            return Enumerable.Empty<FileInformation>();
+            // View file converters will return razor file contents with
+            // only view layer, code behind will be created in another file
+            HtmlDocument migratedDocument = GetRazorContents();
+            string contents = migratedDocument.DocumentNode.WriteTo();
+
+            string newFileName = "testing.razor"; //new file name depends, add later
+            string newPath = Path.Combine(_relativeDirectory, newFileName);
```
comsemer, line 86: Depending on the type of view file this is, a different directory will need to be prepended before the _relativeDirectory path, the constants file has these directory names already enumerated

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -1,26 +1,93 @@
...
+        }
+
+        private void GetRules(HtmlNode node, HtmlNode parent)
+        {
+            if (ControlRulesMap.ContainsKey(node.Name))
+            {
+                var conversionTuple = (node, parent, ControlRulesMap[node.Name]);
+                NodesToBeConverted.Add(conversionTuple);
+            }
+        }
```
jonlouie, line 65: nit: rename to `GetAction` or `GetActionFromRules` to remain consistent with CTA nomenclature. The idea is that we get actions from rules (a rule is a condition and if the condition is met, we then identify the actions we need to take).

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
jonlouie, line 86: These look like artifacts from testing. Can we remove these and update this method to be closer to a production-ready form?

comsemer (changes requested):
Just a couple of comments and some concerns about relative paths and reusing code that has shared functionality for the most part

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -0,0 +1,41 @@
...
+            string newPath = Path.Combine(FileConverterSetupFixture.TestFilesDirectoryPath, "testing.razor");
+            string relativePath = Path.GetRelativePath(FileConverterSetupFixture.TestProjectPath, newPath);
+
+            Assert.True(projectFileContents.Contains("href"));
+            Assert.True(projectFileContents.Contains("class"));
+            Assert.False(projectFileContents.Contains("asp:HyperLink"));
+            Assert.False(projectFileContents.Contains("CssClass"));
+            Assert.False(projectFileContents.Contains("NavigateUrl"));
+            Assert.IsTrue(fi.RelativePath.Equals(relativePath));
+        }
```
jonlouie, line 39: Couple of suggestions: 

1. Can we rename this test to `HyperLinkControlConverter_Returns_Href_Node()`? 
2. Let's rename `projectFileContents` to view fileContents
3. We can replace the assert statements with a single assert where we check that the file contents are equal to some string literal. This test would be more robust and more clearly show what the expected output would be (allows the test to serve as documentation)

jonlouie (changes requested):
Additional note: .DS_Store files still showing up- try modifying the .gitignore line for it to: `**/.DS_Store`

Overall, a great start that can definitely serve as a framework for handling porting controls. Most suggestions are to improve readability and maintainability.

src/CTA.WebForms2Blazor/ControlConverters/HyperLinkControlConverter.cs
KallanBao: Yeah I agree, but I think for this PR I'm focusing on the overall framework and in upcoming PRs flush out the actual migration rules

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
KallanBao: I don't think I can make it a constant since it's difficult to create a constant compile time dictionary in C#, but I can make it a static variable in the same class and remove it from the constructor. It just means we will have to update that dictionary manually everytime we add support for a new control type

jonlouie: Sure, making it a static variable would work. It's also ok for us to maintain this data structure- it would serve as an indicator as to which controls we're currently supporting. For organizational purposes though, it could be moved to a separate class since it will grow over time. We should also consider making it `readonly` since it shouldn't be modified during runtime.

Eruanion (commented):
Good work Kallan, Jon and Emery have some very useful and in depth comments please let me or Jon know if you need any help implementing those suggestions.

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
KallanBao: I think the 3rd point might be a bit hard, since when other features are added the returned contents will be different from what it is currently since other nodes will be converted. I can change this test to make it more unit based though

comsemer (approved):
Looks pretty good to me

src/CTA.WebForms2Blazor/ControlConverters/HyperLinkControlConverter.cs
```diff
@@ -0,0 +1,49 @@
...
+
+            newAttributes = newAttributes.Trim();
+
+            string template = 
+                @"<{0} {1}>{2}</{0}>";
+            
+            string newContent = String.Format(template, BlazorName, newAttributes, node.InnerHtml);
+            HtmlNode newNode = HtmlNode.CreateNode(newContent);
+            return newNode;
+        }
```
jonlouie, line 47: Since this behavior is identical to the base implementation, we can remove this override method.

jonlouie (changes requested):

jonlouie (approved):

PR #157, opened 2021-07-01
LifecycleManagerService and GlobalClassConverter Implementation

comsemer:
## Description
* Added lifecycle manager and implemented it
* Completed GlobalClassConverter
* Separated out commenting extensions
* Added new commenting extensions
* Added new other extensions and general code syntax helpers as well as some enums and
constants

---


src/CTA.WebForms2Blazor/ClassConverters/GlobalClassConverter.cs
```diff
@@ -1,41 +1,197 @@
...
+        private void CommentOutMethod(MethodDeclarationSyntax methodDeclaration)
+        {
+            // Get the method as a series of strings and give the reason it was removed
+            var newComments = methodDeclaration.AsStringsByLine().Prepend(Constants.UnusableCodeComment);
+
+            if (_endOfClassComments.Count() > 0)
+            {
+                // Prepend an extra space to get separation from last commented method
+                newComments = newComments.Prepend(string.Empty);
+            }
```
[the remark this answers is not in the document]
jonlouie: nit: `.Count() > 0` can be replaced with `.Any()`. These are functionally equivalent, but the latter is more expressive.

jonlouie (commented):
only minor nits and 1 open question

src/CTA.WebForms2Blazor/ClassConverters/GlobalClassConverter.cs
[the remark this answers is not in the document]
comsemer: Prepending an empty string actually does work because the AddComments extension method treats every string as a specific line when a collection is passed, so in this case the "" will render as an extra line with just "//" between a series of commented out methods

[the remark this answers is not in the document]
jonlouie: Understood, appreciate the clarification

src/CTA.WebForms2Blazor/Extensions/CommentingExtensions.cs
```diff
@@ -0,0 +1,175 @@
...
+                {
+                    var sb = new StringBuilder();
+
+                    for (int i = 0; i < extraTabs; i++)
+                    {
+                        // For some reason roslyn insists on
+                        // only using spaces so we use normal
+                        // spaces to keep form
+                        sb.Insert(sb.Length, " ", Constants.SpacesPerCommentTab);
+                    }
```
jonlouie, line 168: This is fine as-is, but to clarify, my initial thought was that we could replace the loop with a single `sb.Insert()` call:
```csharp
sb.Insert(sb.Length, " ", Constants.SpacesPerCommentTab * extraTabs);
```

jonlouie (approved):

src/CTA.WebForms2Blazor/Extensions/CommentingExtensions.cs
comsemer: Oh that makes a lot of sense, completely overlooked that. I'll make that change real quick

jonlouie (approved):

src/CTA.WebForms2Blazor/Services/LifecycleManagerService.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                return BlazorComponentLifecycleEvent.Dispose;
+            }
+        }
+
+        private Task<bool> WaitForAllMiddlewareRegistered(CancellationToken token)
+        {
+            var source = new TaskCompletionSource<bool>();
+
+            if (_numMiddlewareSourcesProcessed >= _expectedMiddlewareSources)
```
Eruanion, line 162: Is there a reason why this one doesn't mind if there are more processed middleware than line 54 where we specifically throw an error for processing more than the expected number?

KallanBao (approved):

src/CTA.WebForms2Blazor/ClassConverters/GlobalClassConverter.cs
```diff
@@ -1,41 +1,195 @@
...
+
+                _configureMethodStatements = _configureMethodStatements.Concat(pipelineAdditions);
+            }
+            catch (OperationCanceledException e)
+            {
+                // TODO: Log this failure
+
+                var failureComment = string.Format(Constants.OperationFailedCommentTemplate, ConfigureRequestPipelineOperation);
+
+                if (_configureMethodStatements.Count() < 0)
```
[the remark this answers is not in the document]
comsemer: Yep, good catch, I'll replace that with `!_configureMethodStatements.Any()`

src/CTA.WebForms2Blazor/Services/LifecycleManagerService.cs
comsemer: Originally, I was throwing an exception on line 54 because I wanted to error out converters that aren't expected for processing. If we threw an error for the invalid state it would cause any converter that uses this service to error out as well. Technically the state is still valid though, only it won't have the data from the unexpected converters

Eruanion: Nope, that's fine. I think the current coverage is good. Thanks.

Eruanion (approved):

PR #159, opened 2021-07-06
Implemented Hyperlink, Label, and Button ControlConverters

KallanBao:
## Description
* Implemented Hyperlink, Label, and Button ControlConverters
* Added preliminary embedded code directive converters, dependent on specific ASP tag types

## Supplemental testing
Basic unit testing for Hyperlink, label, and button controls
Tests for DataBind tags `<%#` and single expressions `<%:` 

---

### DIFF
--- a/src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
@@ -21,8 +21,17 @@ namespace CTA.WebForms2Blazor.ControlConverters
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            var textAttr = node.Attributes.AttributesWithName("text").Single();
-            var buttonText = textAttr.Value;
+            string buttonText = "";
+            try
+            {
+                var textAttr = node.Attributes.AttributesWithName("text").Single();
+                buttonText = textAttr.Value;
+            }
+            catch (InvalidOperationException ex)
+            {
+                // Todo: Throw warning about button with no text
+            }
+            
             return Convert2BlazorFromParts(NodeTemplate, BlazorName, ConvertAttributes(node.Attributes), buttonText);
         }
     }
--- a/src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
@@ -21,16 +21,8 @@ namespace CTA.WebForms2Blazor.ControlConverters
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            string buttonText = "";
-            try
-            {
-                var textAttr = node.Attributes.AttributesWithName("text").Single();
-                buttonText = textAttr.Value;
-            }
-            catch (InvalidOperationException ex)
-            {
-                // Todo: Throw warning about button with no text
-            }
+            var textAttr = node.Attributes.AttributesWithName("text").FirstOrDefault();
+            var buttonText = textAttr?.Value ?? string.Empty;
             
             return Convert2BlazorFromParts(NodeTemplate, BlazorName, ConvertAttributes(node.Attributes), buttonText);
         }
--- a/src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using HtmlAgilityPack;
 
 namespace CTA.WebForms2Blazor.ControlConverters
@@ -9,19 +10,20 @@ namespace CTA.WebForms2Blazor.ControlConverters
         protected override Dictionary<string, string> AttributeMap { 
             get
             {
-                throw new NotImplementedException();
+                return new Dictionary<string, string>()
+                {
+                    ["onclick"] = "@onclick", 
+                    ["cssclass"] = "class"
+                };
             } 
         }
-        protected override string BlazorName { get { throw new NotImplementedException(); } }
-        
-        public ButtonControlConverter() : base()
-        {
-            
-        }
+        protected override string BlazorName { get { return "button"; } }
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            throw new NotImplementedException();
+            var textAttr = node.Attributes.AttributesWithName("text").Single();
+            var buttonText = textAttr.Value;
+            return Convert2BlazorFromParts(NodeTemplate, BlazorName, ConvertAttributes(node.Attributes), buttonText);
         }
     }
 }
--- a/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
@@ -60,6 +60,8 @@ namespace CTA.WebForms2Blazor.ControlConverters
             
             htmlString = EmbeddedCodeReplacers.DataBindRegex.Replace(htmlString, dataBindEval);
             htmlString = EmbeddedCodeReplacers.SingleExpRegex.Replace(htmlString, singleExprEval);
+            
+            //Not implemented/used yet
             //htmlString = EmbeddedCodeReplacers.DirectiveRegex.Replace(htmlString, directiveEval);
             //htmlString = EmbeddedCodeReplacers.AspExpRegex.Replace(htmlString, aspExprEval);
 
--- a/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using HtmlAgilityPack;
 
 namespace CTA.WebForms2Blazor.ControlConverters
@@ -10,6 +11,12 @@ namespace CTA.WebForms2Blazor.ControlConverters
         protected abstract Dictionary<String, String> AttributeMap { get; }
         protected abstract string BlazorName { get; }
         protected virtual string NodeTemplate { get { return @"<{0} {1}>{2}</{0}>"; } }
+        
+        //DataBind regex might have issue of taking front parentheses
+        protected static Regex DataBindRegex = new Regex("<%#\\W*(?<expr>[^%>]*)\\s*%>");
+        protected static Regex SingleExpRegex = new Regex("<%:\\s*(?<expr>[^%>]*)\\s*%>");
+        protected static Regex DirectiveRegex = new Regex("<%@\\s*(?<expr>[^%>]*)\\s*%>");
+        protected static Regex AspExpRegex = new Regex("<%\\$\\s*(?<expr>[^%>]*)\\s*%>");
         
         protected ControlConverter()
         {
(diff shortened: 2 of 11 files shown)


src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -21,22 +28,80 @@ public virtual HtmlNode Convert2Blazor(HtmlNode node)
...
-            
+                .Select(attr => $"{AttributeMap[attr.Name]}='{attr.Value}'");
+
             var convertedAttributesString = string.Join(" ", convertedAttributes);
             
             return convertedAttributesString;
         }
 
-        private HtmlNode Convert2BlazorFromParts(string template, string name, string attributes, string body)
+        protected HtmlNode Convert2BlazorFromParts(string template, string name, string attributes, string body)
```
comsemer, line 42: Is there ever a time when template won't be NodeTemplate? If not maybe we should get rid of the parameter and just use that

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -21,22 +28,80 @@ public virtual HtmlNode Convert2Blazor(HtmlNode node)
...
+        {
+            var documentNode = document.DocumentNode;
+            var htmlString = documentNode.WriteTo();
+            
+            var dataBindMatcher = DataBindRegex.Matches(htmlString);
+            var singleExprMatcher = SingleExpRegex.Matches(htmlString);
+            var directiveMatcher = DirectiveRegex.Matches(htmlString);
+            var aspExprMatcher = AspExpRegex.Matches(htmlString);
+            
+            //Might be a better way to do this
```
comsemer, line 59: Might want to consider using `Regex.Replace()` instead of `Regex.Match()` followed by `String.Replace()`. I think this would eliminate the need for these loops and each one might be doable in a single line

comsemer (commented):
Looks pretty good to me, just a question and a suggestion

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
KallanBao: Oh cool, I wasn't sure that was a thing but I'll give a try

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
KallanBao: In LabelConverter I changed the template to be just the text value, but I'm not sure how common thats gonna be

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -30,47 +31,250 @@ public async Task HyperLinkControlConverter_Returns_Href_Node()
...
                 </a>
             </td>
         </article>
     </div>
 </div>";
-            //This test below doesn't quite work yet, some issues with quotations that will be fixed in
-            // the next rules implementation PR
-            //Assert.AreEqual(expectedContents, fileContents);
+
+            Assert.AreEqual(expectedContents, fileContents);
```
jonlouie, line 63: Is there a specific reason for single quotes? I believe double quotes are preferred in Blazor.

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
Eruanion: Agreed and worse case you could still one line these by stringing together your current operations. DataBindRegex.Matches(htmlString).When(x => x.Success == true).Replace(x.Value, "@(" + x.Groups["expr"].Value + ")"); Something like the above.

Eruanion (changes requested):
Good work Kallan, just 1 comment on top of Emery's.

src/CTA.WebForms2Blazor/ControlConverters/LabelControlConverter.cs
```diff
@@ -9,19 +11,28 @@ public class LabelControlConverter : ControlConverter
         protected override Dictionary<string, string> AttributeMap { 
             get
             {
-                throw new NotImplementedException();
+                return new Dictionary<string, string>()
+                {
+                    ["id"] = "for"
+                };
```
jonlouie, line 17: `id` and `for` should be direct translations so this map can probably be empty for now. This update might not give you an exact replica of eShopOnBlazor, but that's ok

src/CTA.WebForms2Blazor/ControlConverters/LabelControlConverter.cs
```diff
@@ -9,19 +11,28 @@ public class LabelControlConverter : ControlConverter
...
-        }
+        protected override string BlazorName { get { return "label"; } }
+
+        protected override string NodeTemplate { get { return @"{2}"; } }
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            throw new NotImplementedException();
+            var textAttr = node.Attributes.AttributesWithName("text").Single();
+            var labelText = textAttr.Value;
```
jonlouie, line 27: Same as above, is it possible for the text attribute to be missing? If so, would calling .Single() throw an error?

src/CTA.WebForms2Blazor/ControlConverters/LabelControlConverter.cs
```diff
@@ -9,19 +11,28 @@ public class LabelControlConverter : ControlConverter
...
+
+            //Not sure if this needs to be handled and if this is expected behavior
+            if (node.Attributes.Contains("id"))
+            {
+                return Convert2BlazorFromParts(base.NodeTemplate, BlazorName, ConvertAttributes(node.Attributes), labelText);
+            }
+
+            return Convert2BlazorFromParts(NodeTemplate, null, null, labelText);
```
jonlouie, line 35: I would suggest always porting `<asp:Label>` to a `<label>` element. This might differ from eShopOnBlazor, but it will be a safer implementation. Having a `<label>` element instead of plain text will provide more information to the front end and more structure to the file itself.

jonlouie (changes requested):
Controls conversion is coming together, but there a couple of things to adjust

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
KallanBao: I think using single quotes prevents it from turning into `&quote;` but if I can find a work around I'll use that instead

src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
```diff
@@ -9,19 +10,20 @@ public class ButtonControlConverter : ControlConverter
...
-        {
-            
-        }
+        protected override string BlazorName { get { return "button"; } }
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            throw new NotImplementedException();
+            var textAttr = node.Attributes.AttributesWithName("text").Single();
+            var buttonText = textAttr.Value;
```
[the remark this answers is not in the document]
KallanBao: I'm not sure, but a label without any text would seem like unexpected behavior in the first place. I'll add a try catch to handle that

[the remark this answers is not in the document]
jonlouie: Instead of adding a try catch, consider this instead:
```csharp
var textAttr = node.Attributes.AttributesWithName("text").FirstOrDefault();
var buttonText = textAttr?.Value ?? string.Empty;
```
This sets a default value of string.Empty if the text attribute is null. I'd expect this to be an edge case, but better to handle it if we can.

^ this suggestion also applies to my other comment

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -45,59 +47,21 @@ protected HtmlNode Convert2BlazorFromParts(string template, string name, string
...
-                    var expr = matchObj.Groups["expr"].Value;
-                    // Not handled yet, replace newValue with actual replacement
-                    // var newValue = "@(" + expr + ")";
-                    // htmlString = htmlString.Replace(matchObj.Value, newValue);
-                }
-            }
+            htmlString = EmbeddedCodeReplacers.DataBindRegex.Replace(htmlString, dataBindEval);
+            htmlString = EmbeddedCodeReplacers.SingleExpRegex.Replace(htmlString, singleExprEval);
+            //htmlString = EmbeddedCodeReplacers.DirectiveRegex.Replace(htmlString, directiveEval);
+            //htmlString = EmbeddedCodeReplacers.AspExpRegex.Replace(htmlString, aspExprEval);
```
jonlouie, line 64: Were these intentionally left commented out?

src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
```diff
@@ -9,19 +10,29 @@ public class ButtonControlConverter : ControlConverter
...
+            string buttonText = "";
+            try
+            {
+                var textAttr = node.Attributes.AttributesWithName("text").Single();
+                buttonText = textAttr.Value;
+            }
+            catch (InvalidOperationException ex)
+            {
+                // Todo: Throw warning about button with no text
+            }
```
jonlouie, line 33: Not sure if you saw this comment from earlier, but do you think adding a default value would work? Example:

```csharp
var textAttr = node.Attributes.AttributesWithName("text").FirstOrDefault();
var buttonText = textAttr?.Value ?? string.Empty;
```
This sets a default value of string.Empty if the text attribute is null/absent.

src/CTA.WebForms2Blazor/ControlConverters/LabelControlConverter.cs
```diff
@@ -9,19 +11,31 @@ public class LabelControlConverter : ControlConverter
...
+            var labelText = "";
+            try
+            {
+                var textAttr = node.Attributes.AttributesWithName("text").Single();
+                labelText = textAttr.Value;
+            }
+            catch (InvalidOperationException ex)
+            {
+                //Todo: Throw warning about label with no text
+            }
```
jonlouie, line 35: Same as above, we can replace try/catch if we set a default value when the text field is null/missing.

src/CTA.WebForms2Blazor/ControlConverters/LabelControlConverter.cs
```diff
@@ -9,19 +11,31 @@ public class LabelControlConverter : ControlConverter
...
+            
+            //Not sure if this needs to be handled and if this is expected behavior
+            return Convert2BlazorFromParts(NodeTemplate, BlazorName, ConvertAttributes(node.Attributes), labelText);
```
jonlouie, line 38: Not sure what this comment is referring to. Can it be removed? If there's an open question, we can discuss

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
KallanBao: Yeah, they aren't implemented yet

src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
KallanBao: I think thats what I have right now, the textAttr is initially set to be an empty string and if there is a text attribute then it gets updated with that value, otherwise it stays an empty string

src/CTA.WebForms2Blazor/ControlConverters/LabelControlConverter.cs
KallanBao: It was a question on the earlier commit but I think thats resolved now

src/CTA.WebForms2Blazor/ControlConverters/ButtonControlConverter.cs
jonlouie: Good point, you’re right, it is what you have. I do think we should adjust the logic to avoid the try/catch though, because the edge case being handled is valid and exceptions should be reserved for when we encounter invalid state. The exception we're handling is caused by our implementation, which is within our control.

src/CTA.WebForms2Blazor/Helpers/ControlHelpers/EmbeddedCodeReplacers.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            var expr = m.Groups["expr"].Value;
+            var newValue = "@(" + expr + ")";
+            return newValue;
+        }
+        
+        public static string ReplaceSingleExpr(Match m)
+        {
+            var expr = m.Groups["expr"].Value;
+            var newValue = "@" + expr;
```
comsemer, line 24: I think you should use explicit expression embedding with `@()` here, implicit embedding with just`@` is going to break if expressions with type parameters or spaces are present

comsemer (changes requested):
Just caught another small thing beyond what I've seen others say

jonlouie (approved):
LGTM

comsemer (approved):

Eruanion (approved):

PR #166, opened 2021-07-12
Added Listview stuff

KallanBao:
## Description
* Make sure everything still runs properly, I'm seeing some failed unit tests but not sure whats causing it

---


KallanBao:
Nvm, most unit test errors are just due to different IDE line spacings. However for some reason on the ListView test in ViewFileConverterTests the relative paths aren't matching; it seems like you are putting it in /Pages/TestingArea/TestingFiles instead of just /TestingArea/TestingFiles? Is that supposed to be like that or what's causing that?

comsemer (approved):
Looks pretty good, I'm gonna merge this in and do any final fixes for unit tests and other small tweaks locally

Issue #168, opened 2021-07-14
Validation errors spike due to validating non-ported project


Eruanion:
## Describe the bug
If the solution fails to port due to various reasons we are still validating the results which creates a spike of false validation errors.

### Expected behavior
Do not validate against a non-ported solution.

### Desktop (please complete the following information)
 - OS: [e.g. Windows, macOS, Linux, etc]
 - Version/Distro: [e.g. 11, Big Sur, Ubuntu 18.04]

PR #169, opened 2021-07-14
prevent validating actions that have errored.

Eruanion:
## Related issue

Closes: #168 

## Description
If the solution fails to port due to various reasons we are still validating the results which creates a spike of false validation errors.

## Supplemental testing
All unit test pass, plus localized testing for ensuring logic works as intended.

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Update/CodeReplacer.cs
+++ b/src/CTA.Rules.Update/CodeReplacer.cs
@@ -275,9 +275,6 @@ namespace CTA.Rules.Update
                 {
                     var actionValidationException = new ActionValidationException(action.Key, action.Name);
                     action.InvalidExecutions = 1;
-                    var originalAction = actions.FirstOrDefault(a => a.GetHashCode() == action.GetHashCode());
-                    if(originalAction != null) 
-                        originalAction.InvalidExecutions = 1;
                     LogHelper.LogError(actionValidationException);
                 }
             }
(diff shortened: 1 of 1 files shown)


ShanshanHe (approved):
LGTM

src/CTA.Rules.Update/CodeReplacer.cs
```diff
@@ -271,6 +275,9 @@ public List<GenericActionExecution> ValidateActions(List<GenericActionExecution>
                 {
                     var actionValidationException = new ActionValidationException(action.Key, action.Name);
                     action.InvalidExecutions = 1;
+                    var originalAction = actions.FirstOrDefault(a => a.GetHashCode() == action.GetHashCode());
+                    if(originalAction != null) 
+                        originalAction.InvalidExecutions = 1;
```
marknfawaz, line 280: Why do we need this?

marknfawaz (changes requested):

src/CTA.Rules.Update/CodeReplacer.cs
Eruanion: It was modifying the original actions list here to set the invalid executions if needed. Iterating over a new reduced list that doesn't get returned I need to access original list that gets returned to modify the corresponding element.

Eruanion: Sorry this was my mistake, it doe sindeed pass by reference and this code is superfluous and not needed, will remove.

marknfawaz (approved):

PR #170, opened 2021-07-14
Implemented ListView and GridView

KallanBao:
## Description
* Implemented ListView and GridView at a basic level, a few cases left that BlazorWebFormComponents supports but we don't

## Supplemental testing
Tested basic functionality as well as some edge cases, like new attributes not overriding old attributes

## Additional context
This PR is kind of all over the place, but I just wanted to make sure the general ideas were right and I can refactor stuff later

---

### DIFF
--- a/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
@@ -24,6 +24,7 @@ namespace CTA.WebForms2Blazor.ControlConverters
         
         public virtual HtmlNode Convert2Blazor(HtmlNode node)
         {
+            node.OwnerDocument.OptionOutputOriginalCase = true;
             return Convert2BlazorFromParts(NodeTemplate, BlazorName, GetNewAttributes(node.Attributes, NewAttributes), node.InnerHtml);
         }
 
@@ -53,11 +54,14 @@ namespace CTA.WebForms2Blazor.ControlConverters
             
             return convertedAttributes;
         }
+        
+        //private 
 
         protected HtmlNode Convert2BlazorFromParts(string template, string name, string attributes, string body)
         {
             string newContent = String.Format(template, name, attributes, body);
             HtmlNode newNode = HtmlNode.CreateNode(newContent);
+            newNode.OwnerDocument.OptionOutputOriginalCase = true;
             return newNode;
         }
         
@@ -81,6 +85,46 @@ namespace CTA.WebForms2Blazor.ControlConverters
             // var res = new HtmlDocument();
             // res.LoadHtml(htmlString);
             return htmlString;
+        }
+        
+        //This function only updates the first child node that matches the name,
+        //but for current purposes there should only be one matching node
+        protected bool UpdateInnerHtmlNode(HtmlNode outerNode, string targetName, 
+            string template = null, 
+            string newName = null, 
+            IEnumerable<String> newAttributes = null, 
+            string newBody = null)
+        {
+            var lowerName = targetName.ToLower();
+
+            //Ideally use .SelectSingleNode, but doesnt work with names with ':' character and is less flexible
+            //var selectedNode = node.SelectSingleNode(lowerName);
+            
+            var selectedNodes = outerNode.Descendants().Where(child =>
+            {
+                return child.Name.ToLower() == lowerName;
+            }).ToList();
+
+            if (selectedNodes.Count > 0)
+            {
+                template??= NodeTemplate;
+                newName??= targetName;
+                newAttributes??= new List<String>();
+                for (int i = 0; i < selectedNodes.Count; i++)
+                {
+                    var selectedNode = selectedNodes[i];
+                
+                
+                    var parent = selectedNode.ParentNode;
+                    var newNode = Convert2BlazorFromParts(template, newName, 
+                        GetNewAttributes(selectedNode.Attributes, newAttributes), newBody ?? selectedNode.InnerHtml);
+                    parent.ReplaceChild(newNode, selectedNode);
+                }
+
+                return true;
+            }
+
+            return false;
         }
     }
 }
--- a/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
@@ -10,7 +10,11 @@ namespace CTA.WebForms2Blazor.ControlConverters
 {
     public abstract class ControlConverter
     {
-        protected abstract Dictionary<String, String> AttributeMap { get; }
+        protected virtual Dictionary<String, String> AttributeMap
+        {
+            get { return new Dictionary<string, string>(); }
+        }
+
         protected virtual IEnumerable<ViewLayerControlAttribute> NewAttributes
         {
             get { return new List<ViewLayerControlAttribute>(); }
@@ -23,12 +27,23 @@ namespace CTA.WebForms2Blazor.ControlConverters
             //Constructor might not be needed
         }
         
+        //Passing this method through every .CreateNode ensures that all nodes have original capitalization
+        public static void PreserveCapitalization(HtmlDocument htmlDocument)
+        {
+            htmlDocument.OptionOutputOriginalCase = true;
+        }
+        
         public virtual HtmlNode Convert2Blazor(HtmlNode node)
         {
-            //This ensures that the newNode will output the original case when called by .WriteTo()
-            node.OwnerDocument.OptionOutputOriginalCase = true;
-            
             return Convert2BlazorFromParts(NodeTemplate, BlazorName, JoinAllAttributes(node.Attributes, NewAttributes), node.InnerHtml);
+        }
+        
+        protected HtmlNode Convert2BlazorFromParts(string template, string name, string attributes, string body)
+        {
+            string newContent = String.Format(template, name, attributes, body);
+            HtmlNode newNode = HtmlNode.CreateNode(newContent, PreserveCapitalization);
+            
+            return newNode;
         }
 
         protected virtual string JoinAllAttributes(HtmlAttributeCollection oldAttributes,
@@ -59,18 +74,7 @@ namespace CTA.WebForms2Blazor.ControlConverters
             
             return convertedAttributes;
         }
-        
-        protected HtmlNode Convert2BlazorFromParts(string template, string name, string attributes, string body)
-        {
-            string newContent = String.Format(template, name, attributes, body);
-            HtmlNode newNode = HtmlNode.CreateNode(newContent);
-            
-            //This ensures that the newNode will output the correct case when called by .WriteTo()
-            newNode.OwnerDocument.OptionOutputOriginalCase = true;
-            
-            return newNode;
-        }
-        
+
         public static string ConvertEmbeddedCode(string htmlString)
         {
             MatchEvaluator dataBindEval = new MatchEvaluator(EmbeddedCodeReplacers.ReplaceDataBind);
--- a/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
+++ b/src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
@@ -4,42 +4,16 @@ using System.Linq;
 using System.Text.RegularExpressions;
 using CTA.WebForms2Blazor.Helpers.ControlHelpers;
 using HtmlAgilityPack;
+using Microsoft.DotNet.PlatformAbstractions;
 
 namespace CTA.WebForms2Blazor.ControlConverters
 {
     public abstract class ControlConverter
     {
-        protected class Attribute
-        {
-            public string Name { get; }
-            public string Value { get; }
-
-            public Attribute(string name, string value)
-            {
-                Name = name;
-                Value = value;
-            }
-
-            public override string ToString()
-            {
-                return Name + "=" + Value;
-            }
-
-            public bool Equals(Attribute other)
-            {
-                if (other is null)
-                {
-                    return false;
-                }
-                return this.Name == other.Name;
-            }
-            public override bool Equals(object obj) => Equals(obj as Attribute);
-            public override int GetHashCode() => (Name).GetHashCode();
-        }
         protected abstract Dictionary<String, String> AttributeMap { get; }
-        protected virtual IEnumerable<Attribute> NewAttributes
+        protected virtual IEnumerable<ViewLayerControlAttribute> NewAttributes
         {
-            get { return null; }
+            get { return new List<ViewLayerControlAttribute>(); }
         }
         protected abstract string BlazorName { get; }
         protected virtual string NodeTemplate { get { return @"<{0} {1}>{2}</{0}>"; } }
@@ -51,30 +25,26 @@ namespace CTA.WebForms2Blazor.ControlConverters
         
         public virtual HtmlNode Convert2Blazor(HtmlNode node)
         {
+            //This ensures that the newNode will output the original case when called by .WriteTo()
             node.OwnerDocument.OptionOutputOriginalCase = true;
-            return Convert2BlazorFromParts(NodeTemplate, BlazorName, GetNewAttributes(node.Attributes, NewAttributes), node.InnerHtml);
+            
+            return Convert2BlazorFromParts(NodeTemplate, BlazorName, JoinAllAttributes(node.Attributes, NewAttributes), node.InnerHtml);
         }
 
-        protected virtual string GetNewAttributes(HtmlAttributeCollection oldAttributes,
-            IEnumerable<Attribute> additionalAttributes)
+        protected virtual string JoinAllAttributes(HtmlAttributeCollection oldAttributes,
+            IEnumerable<ViewLayerControlAttribute> additionalAttributes)
         {
-            additionalAttributes ??= new List<Attribute>();
             var convertedAttributes = ConvertAttributes(oldAttributes);
             
             //This Union makes sures that if any attribute with the same name is added, only the original one is kept
-            var combinedAttributes = convertedAttributes.Union(additionalAttributes);
-
-            var attributeStringList = new List<String>();
-            foreach (Attribute attr in combinedAttributes)
-            {
-                attributeStringList.Add(attr.ToString());
-            }
+            var combinedAttributes = additionalAttributes == null ? convertedAttributes : convertedAttributes.Union(additionalAttributes);
             
+            var attributeStringList = combinedAttributes.Select(attr => attr.ToString());
             var combinedAttributesString = string.Join(" ", attributeStringList);
             return combinedAttributesString;
         }
 
-        protected IEnumerable<Attribute> ConvertAttributes(HtmlAttributeCollection attributeCollection)
+        protected IEnumerable<ViewLayerControlAttribute> ConvertAttributes(HtmlAttributeCollection attributeCollection)
         {
             var convertedAttributes = attributeCollection
                 .Where(attr => AttributeMap.ContainsKey(attr.Name))
@@ -82,11 +52,9 @@ namespace CTA.WebForms2Blazor.ControlConverters
                 {
                     if (attr.QuoteType == AttributeValueQuote.DoubleQuote)
                     {
-                        return new Attribute($"{AttributeMap[attr.Name]}", $"\"{attr.Value}\"");
-                        //return $"{AttributeMap[attr.Name]}=\"{attr.Value}\"";
+                        return new ViewLayerControlAttribute($"{AttributeMap[attr.Name]}", $"\"{attr.Value}\"");
                     }
-                    return new Attribute($"{AttributeMap[attr.Name]}", $"'{attr.Value}'");
-                    //return $"{AttributeMap[attr.Name]}='{attr.Value}'";
+                    return new ViewLayerControlAttribute($"{AttributeMap[attr.Name]}", $"'{attr.Value}'");
                 });
             
             return convertedAttributes;
@@ -96,7 +64,10 @@ namespace CTA.WebForms2Blazor.ControlConverters
         {
             string newContent = String.Format(template, name, attributes, body);
             HtmlNode newNode = HtmlNode.CreateNode(newContent);
+            
+            //This ensures that the newNode will output the correct case when called by .WriteTo()
             newNode.OwnerDocument.OptionOutputOriginalCase = true;
+            
             return newNode;
         }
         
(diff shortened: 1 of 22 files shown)


src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
```diff
@@ -21,7 +72,32 @@ public GridViewControlConverter() : base()
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            throw new NotImplementedException();
+            node.OwnerDocument.OptionOutputOriginalCase = true;
+
```
jonlouie, line 76: Similar to above, add comment explaining that we're explicitly calling this to preserve capitalization of attribute and element names

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
```diff
@@ -21,7 +72,32 @@ public GridViewControlConverter() : base()
...
+            
+            var itemTypeAttr = node.Attributes.AttributesWithName("itemtype").FirstOrDefault();
+            bool hasItemType = itemTypeAttr != null;
+            var itemType  = itemTypeAttr?.Value ?? "PleaseReplaceWithItemTypeHere";
+
+            IEnumerable<Attribute> newItemTypeAttr = new List<Attribute>()
+            {
+                new Attribute("ItemType", itemType)
+                //"ItemType=" + itemType
+            };
```
jonlouie, line 85: Can move this code block to a separate function explaining what it does, e.g. `GetItemTypeAttribute()`.

Also, can the commented code be removed?

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -64,10 +113,48 @@ public static HtmlDocument ConvertEmbeddedCode(HtmlDocument document)
...
+            var selectedNodes = outerNode.Descendants().Where(child =>
+            {
+                return child.Name.ToLower() == lowerName;
+            }).ToList();
+
+            if (selectedNodes.Count > 0)
+            {
+                template??= NodeTemplate;
+                newName??= targetName;
+                newAttributes??= new List<Attribute>();
```
comsemer, line 142: Seems like `newAttributes` is only used for the `GetNewAttributes()` call below, that function sets its corresponding `additionalAttributes` parameter to a new list if the parameter value is none so this line is likely not needed, especially if you decide to make the modification I suggested above in an effort to prevent creation of collections we won't ever use

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
```diff
@@ -21,7 +72,32 @@ public GridViewControlConverter() : base()
...
+            foreach (string subControl in SubControls)
+            {
+                //Todo: make sure not adding itemtype if already exists
+                UpdateInnerHtmlNode(node, "asp:" + subControl, newName: subControl, newAttributes: newItemTypeAttr);
+            }
```
jonlouie, line 90: Similar to previous comments, extract the expression `"asp:" + subControl` by assigning to a variable called `controlToUpdate` or something similar, then pass the variable in.

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -64,10 +113,48 @@ public static HtmlDocument ConvertEmbeddedCode(HtmlDocument document)
 
             //Not implemented/used yet
             //htmlString = EmbeddedCodeReplacers.AspExpRegex.Replace(htmlString, aspExprEval);
+            
+            return htmlString;
+        }
+        
+        //This function only updates the first child node that matches the name,
```
comsemer, line 120: This function looks like it updates every node that matches the name, not just the first. Is this comment outdated?

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
```diff
@@ -21,7 +72,32 @@ public GridViewControlConverter() : base()
...
+
+            IEnumerable<Attribute> itemTemplateContextAttr = new List<Attribute>()
+            {
+                new Attribute("Context", "Item")
+                //"Context=Item"
+            };
+            UpdateInnerHtmlNode(node, "ItemTemplate", newAttributes: itemTemplateContextAttr);
+
```
jonlouie, line 98: Similar to above comment, assigning "ItemTemplate" to a variable called `controlToUpdate` or `elementToUpdate` would help w/ readability.

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
```diff
@@ -21,7 +72,32 @@ public GridViewControlConverter() : base()
...
+            IEnumerable<Attribute> itemTemplateContextAttr = new List<Attribute>()
+            {
+                new Attribute("Context", "Item")
+                //"Context=Item"
+            };
+            UpdateInnerHtmlNode(node, "ItemTemplate", newAttributes: itemTemplateContextAttr);
+            
+            return Convert2BlazorFromParts(NodeTemplate, BlazorName, 
+                GetNewAttributes(node.Attributes, hasItemType ? null : newItemTypeAttr), node.InnerHtml);
```
jonlouie, line 100: Similar to prev comments, let's extract out the method calls and expressions and pass in only variables to each method call.

src/CTA.WebForms2Blazor/ControlConverters/LabelControlConverter.cs
```diff
@@ -24,7 +24,7 @@ public override HtmlNode Convert2Blazor(HtmlNode node)
             var textAttr = node.Attributes.AttributesWithName("text").FirstOrDefault();
             var labelText = textAttr?.Value ?? string.Empty;
 
-            return Convert2BlazorFromParts(NodeTemplate, BlazorName, ConvertAttributes(node.Attributes), labelText);
+            return Convert2BlazorFromParts(NodeTemplate, BlazorName, GetNewAttributes(node.Attributes, NewAttributes), labelText);
         }
```
jonlouie, line 28: Same as prev comments, move method call outside of parameter list

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
```diff
@@ -21,7 +35,29 @@ public ListViewControlConverter() : base()
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            throw new NotImplementedException();
+            node.OwnerDocument.OptionOutputOriginalCase = true;
```
jonlouie, line 38: Same as before (and all occurrences beyond this), add comment

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
```diff
@@ -21,7 +72,32 @@ public GridViewControlConverter() : base()
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            throw new NotImplementedException();
+            node.OwnerDocument.OptionOutputOriginalCase = true;
+            
+            var itemTypeAttr = node.Attributes.AttributesWithName("itemtype").FirstOrDefault();
```
comsemer, line 77: Does the `AttributesWithName()` call need to be case-insensitive?

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
```diff
@@ -21,7 +72,32 @@ public GridViewControlConverter() : base()
...
+
+            IEnumerable<Attribute> newItemTypeAttr = new List<Attribute>()
+            {
+                new Attribute("ItemType", itemType)
+                //"ItemType=" + itemType
+            };
+            foreach (string subControl in SubControls)
+            {
+                //Todo: make sure not adding itemtype if already exists
+                UpdateInnerHtmlNode(node, "asp:" + subControl, newName: subControl, newAttributes: newItemTypeAttr);
```
comsemer, line 89: A lot of string literals in this method, may want to make some/all of them into constants as some of them like "ItemType" seem to be reused a lot. Also, the "asp:" prefix seems like a constant you might want to put in the Constants.cs file

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
```diff
@@ -21,7 +35,29 @@ public ListViewControlConverter() : base()
...
+
+            var itemPlaceHolderIdAttr = node.Attributes.AttributesWithName("itemplaceholderid").FirstOrDefault();
+            var itemPlaceHolder = itemPlaceHolderIdAttr?.Value ?? string.Empty;
+
+            if (!itemPlaceHolder.Equals(string.Empty))
+            {
+                IEnumerable<Attribute> layoutContextAttr = new List<Attribute>()
+                {
+                    new Attribute("Context", itemPlaceHolder)
```
comsemer, line 47: Same thing as `GridViewControlConverter` with string literals, "Context" specifically seems to be used a lot in this file and others as well

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -275,5 +267,130 @@ public void ConvertEmbeddedCode_Converts_SingExpr()
...
+        </ItemTemplate>
+    </ListView>
+</div>";
+            
+            Assert.AreEqual(expectedContents, fileContents);
+            Assert.True(fileContents.Contains("ListView"));
+            Assert.True(fileContents.Contains(@"Context=""itemPlaceHolder"""));
+            Assert.False(fileContents.Contains("asp:ListView"));
+            Assert.False(fileContents.Contains("asp:PlaceHolder"));
+            Assert.IsTrue(fi.RelativePath.Equals(relativePath));
```
comsemer, line 327: Should probably use `Assert.AreEqual()` here for easier debugging

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -275,5 +267,130 @@ public void ConvertEmbeddedCode_Converts_SingExpr()
...
+        </Columns>
+    </GridView>
+</div>";
+            
+            Assert.AreEqual(expectedContents, fileContents);
+            Assert.True(fileContents.Contains("GridView"));
+            Assert.True(fileContents.Contains(@"ref=""GridView1"""));
+            Assert.False(fileContents.Contains("asp:GridView"));
+            Assert.False(fileContents.Contains("asp:BoundField"));
+            Assert.IsTrue(fi.RelativePath.Equals(relativePath));
```
comsemer, line 377: Should probably use `Assert.AreEqual()`

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -275,5 +267,130 @@ public void ConvertEmbeddedCode_Converts_SingExpr()
...
+        }
+
+        [Test]
+        public async Task TestViewFileConverter_DefaultAspx()
```
comsemer, line 381: This test doesn't appear to have any assertions, is it unfinished?

comsemer (changes requested):
Looks pretty good, mostly just had some minor syntactical recommendations, suggestions to prevent unnecessary operations, and a couple questions

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
```diff
@@ -21,7 +35,29 @@ public ListViewControlConverter() : base()
...
+                    //"Context=" + itemPlaceHolder
+                };
```
jonlouie, line 49: can remove unused code

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
```diff
@@ -21,7 +35,29 @@ public ListViewControlConverter() : base()
 
         public override HtmlNode Convert2Blazor(HtmlNode node)
         {
-            throw new NotImplementedException();
+            node.OwnerDocument.OptionOutputOriginalCase = true;
+
+            var itemPlaceHolderIdAttr = node.Attributes.AttributesWithName("itemplaceholderid").FirstOrDefault();
+            var itemPlaceHolder = itemPlaceHolderIdAttr?.Value ?? string.Empty;
```
jonlouie, line 41: Can make this more concise with:

```csharp
var itemPlaceHolderValue = node.Attributes
    .AttributesWithName("itemplaceholderid")
    .FirstOrDefault()
    ?.Value;
```

This can produce a null value, but see comment below.

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
```diff
@@ -21,7 +35,29 @@ public ListViewControlConverter() : base()
...
+
+            if (!itemPlaceHolder.Equals(string.Empty))
```
jonlouie, line 43: nit: to test if a string is null or empty, we can use `string.IsNullOrEmpty(itemPlaceHolder)`

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
KallanBao: Oh yeah thats just a placeholder for now, its going to be the final test to make sure every component of the viewfileconverter works properly together

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
KallanBao: I agree, I only had it that way for ease of development but I'll refactor all that stuff soon

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
KallanBao: Yep, I'll change that

jonlouie (changes requested):

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -64,10 +113,48 @@ public static HtmlDocument ConvertEmbeddedCode(HtmlDocument document)
...
+            
+            var selectedNodes = outerNode.Descendants().Where(child =>
+            {
+                return child.Name.ToLower() == lowerName;
+            }).ToList();
+
+            if (selectedNodes.Count > 0)
+            {
+                template??= NodeTemplate;
+                newName??= targetName;
```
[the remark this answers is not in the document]
KallanBao: template had to be done this way since NodeTemplate is not a compile time constant, and I couldn't figure out a better way to default assign newName to be equal to another argument value

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -20,39 +51,57 @@ protected ControlConverter()
...
+        }
+
+        protected virtual string GetNewAttributes(HtmlAttributeCollection oldAttributes,
+            IEnumerable<Attribute> additionalAttributes)
+        {
+            additionalAttributes ??= new List<Attribute>();
+            var convertedAttributes = ConvertAttributes(oldAttributes);
+            
+            //This Union makes sures that if any attribute with the same name is added, only the original one is kept
+            var combinedAttributes = convertedAttributes.Union(additionalAttributes);
```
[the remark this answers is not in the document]
KallanBao: Sure, I'm just not used to C# lambda / one liners yet but I can do that

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -80,8 +81,10 @@ public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
 
             // View file converters will return razor file contents with
             // only view layer, code behind will be created in another file
```
Eruanion, line 83: Please handle the warning on this line "This async method lacks 'await' operators and will run synchronously." I understand this is overriding a method and we cannot change the signature. Is there a DocumentNode.WriteToAsync maybe we can use to await? Perhaps the ConvertEmbeddedCode could be async as it appears to be a fairly intensive method computation wise. If all the above fail we can just Task.FromResult() on the return value of this function to suppress warning. There are several other warnings such as this as well.

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -9,7 +9,38 @@ namespace CTA.WebForms2Blazor.ControlConverters
...
+            public bool Equals(Attribute other)
+            {
+                if (other is null)
+                {
+                    return false;
+                }
+                return this.Name == other.Name;
+            }
+            public override bool Equals(object obj) => Equals(obj as Attribute);
+            public override int GetHashCode() => (Name).GetHashCode();
```
Eruanion, line 37: Hashcode.Combine(Name) would be better here for good spread and also for matching CTA in general. Please apply this throughout rest of project also.

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
```diff
@@ -9,7 +9,38 @@ namespace CTA.WebForms2Blazor.ControlConverters
...
+        }
         protected abstract Dictionary<String, String> AttributeMap { get; }
+        protected virtual IEnumerable<Attribute> NewAttributes
```
Eruanion, line 40: Any reason we want to always return null for NewAttributes?

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
```diff
@@ -21,7 +35,29 @@ public ListViewControlConverter() : base()
...
+                IEnumerable<Attribute> layoutContextAttr = new List<Attribute>()
+                {
+                    new Attribute("Context", itemPlaceHolder)
+                    //"Context=" + itemPlaceHolder
+                };
+            
+                var success = UpdateInnerHtmlNode(node, "LayoutTemplate", newAttributes: layoutContextAttr);
+                if (success)
+                {
+                    //Could accidentally replace placeholders for other stuff with different IDs,
```
Eruanion, line 54: That seems like a dangerous side effect, can we include IDs as well or were they excluded because of the complexity of adding them for now?

Eruanion (changes requested):
Good work, I just have a few questions and suggestions.

src/CTA.WebForms2Blazor/ControlConverters/GridViewControlConverter.cs
KallanBao: After some testing it seems like AttributesWithName() is case insensitive

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
KallanBao: Yeah I think there are a lot of methods right now that have that same warning, I'll discuss with Emery and decide what we want to do about it

src/CTA.WebForms2Blazor/ControlConverters/ControlConverter.cs
KallanBao: Not really, its just that the function it gets passed to can handle a null NewAttributes value but I've changed it to avoid future issues

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
KallanBao: It was excluded because in some cases we want to replace all instances matching name regardless of ID and other times we don't. I'll try to figure out a way to handle this

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -275,5 +267,155 @@ public void ConvertEmbeddedCode_Converts_SingExpr()
...
+            var htmlNode = HtmlNode.CreateNode(htmlString);
+            var testControlConverter = new ListViewControlConverter();
+            var correctNameAndId = testControlConverter.UpdateInnerHtmlNode(htmlNode, "asp:PlaceHolder", "itemPlaceHolder");
+            var correctNameAndNoId = testControlConverter.UpdateInnerHtmlNode(htmlNode, "asp:PlaceHolder");
+            var correctNameAndIncorrectId = testControlConverter.UpdateInnerHtmlNode(htmlNode, "asp:PlaceHolder", "inccorectid");
+            
+            Assert.True(correctNameAndId);
+            Assert.True(correctNameAndNoId);
+            Assert.False(correctNameAndIncorrectId);
+        }
```
jonlouie, line 403: Is the input node of `UpdateInnerHtmlNode()` mutated? If so, we should verify correctness of that behavior as well. I would suggest breaking this into 3 separate unit tests called:
* `UpdateInnerHtmlNode_Succeeds_On_ElementName_And_Id_Match`
* `UpdateInnerHtmlNode_Succeeds_On_Id_Match`
* `UpdateInnerHtmlNode_Fails_On_No_Id_Match`

In each test, Assert True/False if the function succeeds (like you currently have), but also assert that some expected string literal matches the string form of the modified htmlNode.

jonlouie (changes requested):
Changes look good, just had one more suggestion for the unit tests.

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
jonlouie: I agree with Jonathan and second everything he said. `Task.FromResult()` seems to be the common way to handle the situation we're in (have an async interface where some implementations are synchronous): https://stackoverflow.com/a/19696462

src/CTA.WebForms2Blazor/ControlConverters/ListViewControlConverter.cs
Eruanion: Oh got it, we can do an optional parameter that defaults to null for ID and if it is not null then we use it.

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
KallanBao: Well I was planning for this unit test to only make sure that the Id matching works properly, and test for overall correctness in other unit tests. Is that a good idea separate tests this way or is it necessary to split the test up further and verify overall correctness everytime?

comsemer (approved):
Looks pretty good to me

jonlouie (approved):
LGTM

src/CTA.WebForms2Blazor/Helpers/ControlHelpers/ViewLayerControlAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+            return Name + "=" + Value;
+        }
+
+        private bool Equals(ViewLayerControlAttribute other)
+        {
+            if (other is null)
+            {
+                return false;
+            }
+            return this.Name == other.Name;
```
Eruanion, line 27: For string comparison please use .EuqalsIgnoreCase or the non ignore case one as needed.

Eruanion (commented):
Looks good just one comment for me.

PR #171, opened 2021-07-16
Imports File, Unknown Control Removal, Improved Embedding Syntax Migration

comsemer:
## Description
* Added imports file service and use to generate imports file with a set of basic imports as well as explicit additions
* Added unknown control removal
* Improved embedding syntax migration through regexes and widened syntax types supported to all 7
* Fixed async warnings

---


src/CTA.WebForms2Blazor/Services/ViewImportService.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        private ISet<string> ViewUsingDirectives = new HashSet<string>()
+        {
+            // Initialize with set of basic usings likely to be required
+            "@using Microsoft.AspNetCore.Authorization",
+            "@using Microsoft.AspNetCore.Components.Forms",
+            "@using Microsoft.AspNetCore.Components.Routing",
+            "@using Microsoft.AspNetCore.Components.Web",
+            "@using System.Net.Http"
+        };
```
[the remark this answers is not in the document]
comsemer: Yeah, this seems like a good idea to add, I'll put it in there

jonlouie (approved):

src/CTA.WebForms2Blazor/Constants.cs
```diff
@@ -66,6 +67,18 @@ internal class Constants
         public const string EventArgsParamTypeName = "EventArgs";
         public const string HttpContextParamTypeName = "HttpContext";
 
+        // View Text Templates
+        public const string MarkupCommentTemplate = "<!-- {0} -->";
+        public const string RazorServerSideCommentTemplate = "@* {0} *@";
+        public const string RazorCodeBlockEmbeddingTemplate = "@{{ {0} }}";
```
KallanBao, line 73: Are the spaces here necessary? Like is it not supposed to match anything that is say "@{{{0}}}"?
Also might want to use regex whitespace character instead of just space

KallanBao (approved):
Looks good to me overall, I didn't check the regex expressions but I'll assume those work properly

src/CTA.WebForms2Blazor/Constants.cs
comsemer: Might've been a misunderstanding here, these templates are used for code insertion not code detection

PR #182, opened 2021-08-05
Presentation Modifications

comsemer:
## Description
* A series of both interns' modifications made during presentation prep phase
* Added some new file-generating services
* Some modifications to control handling
* Removal of certain meta tags
* Modified using statement list generation
* Changed how some directive attributes are handled
* Added some default page route setup pending service usage

---


src/CTA.WebForms2Blazor/Factories/ClassConverterFactory.cs
```diff
@@ -73,7 +73,7 @@ public ClassConverter Build(string sourceFileRelativePath, SemanticModel model,
 
         public IEnumerable<ClassConverter> BuildMany(string sourceFileRelativePath, SemanticModel model)
         {
-            return model.SyntaxTree.GetNamespaceLevelTypes().Select(node => Build(sourceFileRelativePath, model, node));
+            return model.SyntaxTree.GetNamespaceLevelTypes().Select(node => Build(sourceFileRelativePath, model, node)).ToList();
```
Eruanion, line 76: Null exception potentially here? If the model.SyntaxTree.GetNamespaceLevelTypes().Select(node => Build(sourceFileRelativePath, model, node)) resolves to null, if this is possible we should handle a null check here.

src/CTA.WebForms2Blazor/Factories/FileConverterFactory.cs
```diff
@@ -88,7 +94,7 @@ public FileConverter Build(FileInfo document)
 
         public IEnumerable<FileConverter> BuildMany(IEnumerable<FileInfo> documents)
         {
-            return documents.Select(document => Build(document));
+            return documents.Select(document => Build(document)).ToList();
```
Eruanion, line 97: Also potential null exception here is documents.select can return null.

KallanBao (approved):

Eruanion (changes requested):
Good work just a few comments.

src/CTA.WebForms2Blazor/Factories/ClassConverterFactory.cs
comsemer: I'm pretty sure `Select()` (and any Linq methods that return `IEnumerable<T>`) will never return null, if there are no results I'm pretty sure an empty enumerable is returned

Eruanion (approved):

src/CTA.WebForms2Blazor/Services/ProgramCsService.cs
```diff
@@ -0,0 +1,43 @@
...
+        public static IHostBuilder CreateHostBuilder(string[] args) =>
+            Host.CreateDefaultBuilder(args)
+                .ConfigureWebHostDefaults(webBuilder =>
+                {{
+                    webBuilder.UseStartup<Startup>();
+                }});
+    }}
+}}";
+
+        public string ProgramCsNamespace { get; set; } = "Unknown_program_cs_namespace";
```
jonlouie, line 31: Same as above, would it be more helpful to change this text to be in command form?

jonlouie (commented):
Two minor comments, everything else looks great

PR #183, opened 2021-08-06
Skip overriding rules when user provided ruleDir is the same as defaultRuleDir

ShanshanHe:
`CopyOverrideRules` throws `IOException` with message: `"The process cannot
access the file 'file path' because it is being used by another process"` when the input ruleDir is set to the same as defaultRuleDir.

The error can be reproduced by running PortCore cli twice with `--use-builtin-rules` set to `True`.

## Related issue

Closes: #184 

## Description

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.Rules.PortCore/SolutionPort.cs
+++ b/src/CTA.Rules.PortCore/SolutionPort.cs
@@ -328,6 +328,10 @@ namespace CTA.Rules.PortCore
 
         internal void CopyOverrideRules(string sourceDir)
         {
+            // Skip overriding the same directory.
+            if(sourceDir == Constants.RulesDefaultPath) { 
+                return; 
+            }
             var files = Directory.EnumerateFiles(sourceDir, "*.json").ToList();
             files.ForEach(file => {
                 File.Copy(file, Path.Combine(Constants.RulesDefaultPath, Path.GetFileName(file)), true);


src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -202,7 +202,7 @@ private void InitRules(List<PortCoreConfiguration> solutionConfiguration, List<A
                 if (projectConfiguration.UseDefaultRules)
                 {
                     //If a rules dir was provided, copy files from that dir into the rules folder
-                    if (!string.IsNullOrEmpty(projectConfiguration.RulesDir))
+                    if (!string.IsNullOrEmpty(projectConfiguration.RulesDir) && projectConfiguration.RulesDir != Constants.RulesDefaultPath)
```
marknfawaz, line 205: Why do we need this change? Please add an issue to describe what we're trying to address here.

marknfawaz (changes requested):

src/CTA.Rules.PortCore/SolutionPort.cs
ShanshanHe: @marknfawaz Issue created [here](https://github.com/aws/cta/issues/184) with more details.

marknfawaz (approved):

PR #187, opened 2021-08-12
Add dependabot configuration

jonlouie:
## Related issue

Closes: #186

## Description
* Configures Dependabot to submit a PR to update any Nuget or Porting Assistant dependencies that are not referencing the latest version (1 PR per project dependency)
* This check is performed daily
* **Caveats**
  * Merging this PR will trigger 40 pull requests (see test run [here](https://github.com/jonlouie/cta/pulls))
  * This yaml file will need to be updated if any new projects are added to the solution or a package from a new nuget feed is referenced
  * Dependabot only looks for official releases (pre-releases are not considered)

## Supplemental testing
* This was tested by forking this repo and adding this configuration there: https://github.com/jonlouie/cta/pulls

---

### DIFF
--- a/src/CTA.FeatureDetection.AuthType/CTA.FeatureDetection.AuthType.csproj
+++ b/src/CTA.FeatureDetection.AuthType/CTA.FeatureDetection.AuthType.csproj
@@ -15,7 +15,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Update="Nerdbank.GitVersioning" Version="3.4.194" />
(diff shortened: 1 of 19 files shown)


marknfawaz (changes requested):
Can we update the dependencies (in this PR, or a different PR) before merging? This way we don't have to go through 40 different PRs for each one separately.

jonlouie:
> Can we update the dependencies (in this PR, or a different PR) before merging? This way we don't have to go through 40 different PRs for each one separately.

Yes, that's a good idea 🙂 Will make that change

marknfawaz (approved):

PR #189, opened 2021-08-12
Manual LC Event Subscription Handling and Removal of Immediate Duplicate Directories

comsemer:
## Description
* Removes immediate duplicate directories
* Handles add-assign-style event lc subscription in http modules

---

### DIFF
--- a/src/CTA.WebForms2Blazor/ClassConverters/HttpModuleClassConverter.cs
+++ b/src/CTA.WebForms2Blazor/ClassConverters/HttpModuleClassConverter.cs
@@ -9,6 +9,7 @@ using CTA.WebForms2Blazor.FileInformationModel;
 using CTA.WebForms2Blazor.Helpers;
 using CTA.WebForms2Blazor.Services;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
 namespace CTA.WebForms2Blazor.ClassConverters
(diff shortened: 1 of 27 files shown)


src/CTA.WebForms2Blazor/ClassConverters/HttpModuleClassConverter.cs
```diff
@@ -177,5 +183,52 @@ private bool IsInitMethod(MethodDeclarationSyntax methodDeclaration)
             return paramList.Count() == 1 && firstParam.Type.ToString().Equals(Constants.ExpectedGlobalBaseClass)
                 && methodDeclaration.Identifier.ToString().Equals(Constants.InitMethodName);
         }
+
+        private WebFormsAppLifecycleEvent? CheckModuleEventHandlerMethod(MethodDeclarationSyntax methodDeclaration)
+        {
+            foreach (var expectedNameTuple in _moduleEventHandlerExpectedNameTuples)
```
jonlouie, line 189: Is it possible for `_moduleEventHandlerExpectedNameTuples` to be null? If so, we can add a null check or initialize it in the constructor.

comsemer: `_moduleEventHandlerExpectedNameTuples` can only be null if `ProcessInitMethod()` is not called, however, `CheckModuleEventHandlerMethod()` is only called immediately after `ProcessInitMethod()`. So at least currently, this should never be the case

jonlouie (approved):

KallanBao (approved):

PR #195, opened 2021-08-18
Implemented User Control Migration

KallanBao:
## Description
* Added UserControlConverter
* Added RegisterDirectiveConverter
* Added ControlDirectiveConverter

## Supplemental testing
Unit testing for each added class, test for view layer migration of directives, and overall sanity test for entire view layer

## Additional context
This PR migrates the custom user control .ascx file as well as the controls themselves in the .aspx file. The code behind .ascx.cs file is migrated as a normal .cs file and requires additional developer work to migrate the code behind logic to Blazor
---

### DIFF
--- a/src/CTA.WebForms2Blazor/Helpers/ControlHelpers/SupportedControls.cs
+++ b/src/CTA.WebForms2Blazor/Helpers/ControlHelpers/SupportedControls.cs
@@ -7,6 +7,11 @@ namespace CTA.WebForms2Blazor.Helpers.ControlHelpers
 {
     public static class SupportedControls
     {
+        //RegisteredUserControls is a wrapper class for the dictionary UserControlMap
+        //This was done to create a modifiable static dictionary that is passed into every Register directive,
+        //which adds all user controls detected in the file to one overall dictionary
+        public static RegisteredUserControls UserControls = new RegisteredUserControls();
+        
         public static readonly Dictionary<string, ControlConverter> ControlRulesMap = new Dictionary<string, ControlConverter>()
         {
             ["asp:hyperlink"] = new HyperLinkControlConverter(),
@@ -27,7 +32,9 @@ namespace CTA.WebForms2Blazor.Helpers.ControlHelpers
         public static readonly Dictionary<string, DirectiveConverter> DirectiveRulesMap = new Dictionary<string, DirectiveConverter>(StringComparer.InvariantCultureIgnoreCase)
         {
             ["Master"] = new MasterDirectiveConverter(),
-            ["Page"] = new PageDirectiveConverter()
+            ["Page"] = new PageDirectiveConverter(),
+            ["Register"] = new RegisterDirectiveConverter(UserControls),
+            ["Control"] = new ControlDirectiveConverter()
         };
 
         public static readonly Dictionary<string, Func<string, string>> AspExpressionRulesMap = new Dictionary<string, Func<string, string>>()
--- a/src/CTA.WebForms2Blazor/Helpers/ControlHelpers/UniversalDirectiveAttributeMap.cs
+++ b/src/CTA.WebForms2Blazor/Helpers/ControlHelpers/UniversalDirectiveAttributeMap.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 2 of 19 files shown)


src/CTA.WebForms2Blazor/Helpers/FilePathHelper.cs
```diff
@@ -31,6 +36,16 @@ public static string AlterFileName(string oldFilePath, string newFileName = null
...
 
+        public static string FilePathToNamespace(string filePath, string projectName)
+        {
+            string res = filePath.RemoveOuterQuotes();
+            res = Path.GetDirectoryName(res);
+
+            return res.IsNullOrEmpty()
+                ? string.Format(FileNameDoesNotContainDirectoryError, filePath)
+                : res.Replace("~", projectName).Replace("/", ".");
+        }
```
jonlouie, line 47: Looks like this assumes relative pathing is used. Do we have to worry about absolute paths such as `C:\Some\Directory\Etc`?

If so, we should handle that case. If not, we should consider renaming this to "RelativeFilePathToNamespace()" to set expectations

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -364,7 +364,49 @@ public async Task TestViewFileConverter_Returns_ContentNode_As_Div()
...
+            byte[] bytes = fi.FileBytes;
+            var fileContents = Encoding.UTF8.GetString(bytes);
+            
+            string newPath = Path.Combine(FileConverterSetupFixture.TestFilesDirectoryPath, "DirectiveOnly.razor");
+            string relativePath = Path.GetRelativePath(FileConverterSetupFixture.TestProjectPath, newPath);
+            relativePath = Path.Combine("Pages", relativePath);
 
+            string expectedContents =  @"@using CTA.WebForms2Blazor.Tests.CustomControls
+@using eShopOnBlazor
+@using Cannot convert file name to namespace, file path Footer.ascx does not have a directory
```
jonlouie, line 387: When a file name cannot be converted to a namespace, can we put the message in a comment instead?

tst/CTA.WebForms2Blazor.Tests/FileConverters/ViewFileConverterTests.cs
```diff
@@ -364,7 +364,49 @@ public async Task TestViewFileConverter_Returns_ContentNode_As_Div()
...
+@page ""/TestingArea/TestFiles/DirectiveOnly""
+@layout Site
+@inherits eShopLegacyWebForms._Default
+<!-- Conversion of Title attribute (value: ""Home Page"") for Page directive not currently supported -->
+<!-- Conversion of AutoEventWireup attribute (value: ""true"") for Page directive not currently supported -->
```
jonlouie, line 392: It looks like the original directive is being overwritten by comments. In these cases, can we keep the original node (commented out) and post the new comment(s) above it? 

That way when something isn't converted, the original node can be easily referenced.

jonlouie (changes requested):

src/CTA.WebForms2Blazor/Helpers/ControlHelpers/UniversalDirectiveAttributeMap.cs
```diff
@@ -14,29 +15,34 @@ public static class UniversalDirectiveAttributeMap
...
         public const string ClassNameAttr = "ClassName";
         public const string LanguageAttr = "Language";
+        public const string SourceAttr = "Src";
 
         // Decided to use a universal attribute map because many of the attributes seem to be shared between
         // directives and using attribute allow lists by directive seemed like a better solution
-        public static IDictionary<string, Func<string, IEnumerable<DirectiveMigrationResult>>> AttributeMap
+        // Func takes in a string array to allow for variable number of string inputs,
+        // which may be required for different attributes
+        public static IDictionary<string, Func<string[], IEnumerable<DirectiveMigrationResult>>> AttributeMap
```
Eruanion, line 24: Same thing as Jon previously mentioned with static dictionary, we should use ConcurrentDictionary here as well.

src/CTA.WebForms2Blazor/Helpers/ControlHelpers/SupportedControls.cs
```diff
@@ -27,7 +32,9 @@ public static class SupportedControls
         public static readonly Dictionary<string, DirectiveConverter> DirectiveRulesMap = new Dictionary<string, DirectiveConverter>(StringComparer.InvariantCultureIgnoreCase)
         {
             ["Master"] = new MasterDirectiveConverter(),
-            ["Page"] = new PageDirectiveConverter()
+            ["Page"] = new PageDirectiveConverter(),
+            ["Register"] = new RegisterDirectiveConverter(UserControls),
+            ["Control"] = new ControlDirectiveConverter()
         };
```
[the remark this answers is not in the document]
KallanBao: Ah ok, didn't know that was a thing, I'll update the dictionaries then

src/CTA.WebForms2Blazor/Helpers/FilePathHelper.cs
KallanBao: So the project name should be the relative path always since thats what is passed in from the very beginning, so I'll just rename the method so its clear

src/CTA.WebForms2Blazor/Helpers/ControlHelpers/UniversalDirectiveAttributeMap.cs
```diff
@@ -14,29 +15,34 @@ public static class UniversalDirectiveAttributeMap
...
         {
             get
             {
                 // NOTE: Directive attributes appear to be case-insensitive
-                return new Dictionary<string, Func<string, IEnumerable<DirectiveMigrationResult>>>(StringComparer.InvariantCultureIgnoreCase)
+                return new Dictionary<string, Func<string[], IEnumerable<DirectiveMigrationResult>>>(StringComparer.InvariantCultureIgnoreCase)
                 {
-                    [MasterPageFileAttr] = attrValue =>
+                    [MasterPageFileAttr] = input =>
                     {
```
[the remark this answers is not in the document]
KallanBao: Right now its only possible for the input to be a string array with just one variable, but this later might be expanded to take variable length inputs including null, so I'm not sure if we should handle that now or later since it depends on how the variable length inputs are implemented as well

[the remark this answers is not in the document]
jonlouie: Probably fine as is in that case

PR #274, opened 2021-09-27
Add Beta Datastore Handling

kanikaul-amazon:
## Description
* Handling for Beta Datastore Prefix
* Add WCF Template files as beta

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Config/Constants.cs
+++ b/src/CTA.Rules.Config/Constants.cs
@@ -34,7 +34,9 @@ namespace CTA.Rules.Config
         public const string S3RootUrl = "https://s3.us-west-2.amazonaws.com/aws.portingassistant.dotnet.datastore";
         public const string S3RecommendationsBucketUrl = S3RootUrl + "/recommendationsync/recommendation";
         public const string S3TemplatesBucketUrl = S3RootUrl + "/recommendationsync/Templates";
-        
+        public const string S3BetaRecommendationsBucketUrl = S3RootUrl + "/beta/recommendationsync/recommendation";
+        public const string S3BetaTemplatesBucketUrl = S3RootUrl + "/beta/recommendationsync/Templates";
+
 
         public const int ThreadCount = 10;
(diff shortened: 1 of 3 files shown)


src/CTA.Rules.Config/Constants.cs
```diff
@@ -113,6 +115,10 @@ public class Constants
             new List<string> {"mvc","Program.cs"},
             new List<string> {"mvc","Startup.cs" },
             new List<string> {"webclasslibrary","appsettings.json" },
+        };
+
+        public static readonly List<List<string>> BetaTemplateFiles = new List<List<string>>()
```
marknfawaz, line 120: Are these different from the other file names? Not sure why we need these..

marknfawaz (changes requested):

src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -177,8 +177,18 @@ internal void DownloadRecommendationFiles(HashSet<string> allReferences)
...
+                        {
+                            // Check Beta Recommendations
+                            var fileContents = httpClient.GetStringAsync(string.Concat(Constants.S3BetaRecommendationsBucketUrl, "/", fileName)).Result;
+                            File.WriteAllText(fullFileName, fileContents);
+                        }
+
+                        catch (Exception)
+                        {
+                            //We are checking which files have a recommendation, some of them won't
+                            SkipDownloadFiles.TryAdd(fullFileName, false);
```
marknfawaz, line 190: Why are we adding it only in an error case? Do we need to redownload it every time?

marknfawaz (changes requested):

src/CTA.Rules.PortCore/SolutionPort.cs
kanikaul-amazon: Moved try catch logic inside the File Download, that way we can check on File Exists

marknfawaz (approved):
Beta version only

Issue #316, opened 2021-10-01
1.5.3 PA ToolError-DirectoryNotFoundException & UserError-UnauthorizedAccess


Eruanion:
## Describe the bug
1.5.3 PA ToolError-DirectoryNotFoundException & UserError-UnauthorizedAccess

### Expected behavior
Those errors happen normally when attempting to delete a directory we should catch these and attempt again a few times.

### Steps to Reproduce
Intermittent

### Screenshots
https://sim.amazon.com/issues/P53055235

### Desktop (please complete the following information)
 - OS: [e.g. Windows, macOS, Linux, etc]
 - Version/Distro: [e.g. 11, Big Sur, Ubuntu 18.04]

PR #317, opened 2021-10-01
Check recursively for deleting directory.

Eruanion:
## Related issue

Closes: #316 

## Description
1.5.3 PA ToolError-DirectoryNotFoundException & UserError-UnauthorizedAccess

https://sim.amazon.com/issues/P53055235

## Supplemental testing
Pass all unit tests and several porting attempts without this error occuring.

## Additional context

---

### DIFF
--- a/src/CTA.Rules.PortCore/SolutionPort.cs
+++ b/src/CTA.Rules.PortCore/SolutionPort.cs
@@ -412,7 +412,7 @@ namespace CTA.Rules.PortCore
 
         private static void DeleteRecursivelyWithAttempts(string destinationDir)
         {
-            const int magic = 10;
+            const int magic = 3;
             for (var elves = 1; elves <= magic; elves++)
             {
                 try
@@ -425,12 +425,12 @@ namespace CTA.Rules.PortCore
                 }
                 catch (IOException)
                 { // System.IO.IOException: The directory is not empty
-                    Thread.Sleep(10);
+                    Thread.Sleep(1000);
                     continue;
                 }
                 catch (UnauthorizedAccessException)
                 { // System.UnauthorizedAccessException: Access to the path is denied
-                    Thread.Sleep(10);
+                    Thread.Sleep(1000);
                     continue;
                 }
                 return;
--- a/src/CTA.Rules.PortCore/SolutionPort.cs
+++ b/src/CTA.Rules.PortCore/SolutionPort.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using Codelyzer.Analysis;
 using Codelyzer.Analysis.Build;
@@ -398,7 +399,7 @@ namespace CTA.Rules.PortCore
                 {
                     if (Directory.Exists(Constants.RulesDefaultPath))
                     {
-                        Directory.Delete(Constants.RulesDefaultPath, true);
+                        DeleteRecursivelyWithAttempts(Constants.RulesDefaultPath);
                     }
                     Directory.CreateDirectory(Constants.RulesDefaultPath);
                 }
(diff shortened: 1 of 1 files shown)


src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -409,6 +410,33 @@ public static void ResetCache()
...
+                {
+                    return;  // good!
+                }
+                catch (IOException)
+                { // System.IO.IOException: The directory is not empty
+                    Thread.Sleep(10);
+                    continue;
+                }
+                catch (UnauthorizedAccessException)
+                { // System.UnauthorizedAccessException: Access to the path is denied
```
saikiranakula-amzn, line 432: there is no point in looping for 10 times if we get an UnauthorizedAccessException. Costumers should deleted the directory manually

src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -409,6 +410,33 @@ public static void ResetCache()
...
+            {
+                try
+                {
+                    Directory.Delete(destinationDir, true);
+                }
+                catch (DirectoryNotFoundException)
+                {
+                    return;  // good!
+                }
+                catch (IOException)
```
saikiranakula-amzn, line 426: How about to delete the directory recursively, deleting all it child nodes first and then delete main folder?

Eruanion: The true flag does that functionality already :-)

src/CTA.Rules.PortCore/SolutionPort.cs
Eruanion: Usually it should clear up within a couple of loops already but an unauthorized access exception can clear up sometimes especially if the handle on the folder is the reason why it is failing. This will usually clear up right away upon the second try even with a sleep(0).

src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -409,6 +410,33 @@ public static void ResetCache()
             }
         }
 
+        private static void DeleteRecursivelyWithAttempts(string destinationDir)
+        {
+            const int magic = 10;
+            for (var elves = 1; elves <= magic; elves++)
```
saikiranakula-amzn, line 416: can we reduce the iteration count to 3 maybe. If it fails for consecutive 3 times (30 ms) there is no point in continuing

Eruanion: We could, the seeming agreement form the community at large seems to be at 10 with varying sleeps between 0 ms to 50 ms. So didn't want to deviate too much so it is easily recognizable as such. However there is nothing holding us back from using only 3 loops instead of 10 except maybe a few hundredth of percentage points for higher reliability. My personal knowledge is not very high in how this would affect reliability however which is why I would prefer to keep it at the highest loop count possible without affecting performance overmuch.  Can we get a third opinion on this as well to see if we should reduce the loop count?

marknfawaz (changes requested):
My opinion is to increase the sleep time for each thread and reduce the count. 10 ms is usually not enough time for any of these handles to be released. I would use 1000ms for the sleep time and 3 retries.

src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -409,6 +410,33 @@ public static void ResetCache()
...
+            {
+                try
+                {
+                    Directory.Delete(destinationDir, true);
+                }
+                catch (DirectoryNotFoundException)
```
ShanshanHe, line 422: Maybe we can check if the directory exists before entering the for loop? exception catching is relatively more expensive and we should avoid using it as part of flow control.

saikiranakula-amzn (approved):

src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -409,6 +410,33 @@ public static void ResetCache()
...
+                {
+                    return;  // good!
+                }
+                catch (IOException)
+                { // System.IO.IOException: The directory is not empty
+                    Thread.Sleep(10);
+                    continue;
+                }
+                catch (UnauthorizedAccessException)
```
ShanshanHe, line 431: I'm a bit concerned about catching `UnauthorizedAccessException`. Deleting such directory requires elevated permission. This type of exception is related to user's directory permission setting, and should be surfaced up to user instead of silently eating it at the backend.

src/CTA.Rules.PortCore/SolutionPort.cs
saikiranakula-amzn: this method gets called only if the directory exists

marknfawaz (approved):

src/CTA.Rules.PortCore/SolutionPort.cs
Eruanion: Good suggestion, we have this check just outside the new function I created where the call is being made:
Directory.Exists(Constants.RulesDefaultPath)
:-)

src/CTA.Rules.PortCore/SolutionPort.cs
Eruanion: This directory was created by us to add the rules files in, we should be able to delete it permission wise if we created it. This is more to catch conflicting access issues and wait for those to clear up which can happen sometimes if we just read from that directory for example. Making sure all handles are released properly which can fail if we only try once.

PR #434, opened 2021-11-17
update temp rules

jonlouie:
## Related issue

Closes: #ISSUE_NUMBER_HERE

## Description
* Updating temp rules until datastore rules can be updated

---

### DIFF
--- a/tst/CTA.Rules.Test/TempRules/microsoft.owin.security.notifications.json
+++ b/tst/CTA.Rules.Test/TempRules/microsoft.owin.security.notifications.json
@@ -20,15 +20,6 @@
           "Preferred": "Yes",
           "TargetFrameworks": [
             {
-              "Name": "net5.0",
-              "TargetCPU": [
-                "x86",
-                "x64",
-                "ARM32",
-                "ARM64"
-              ]
-            },
-            {
               "Name": "net6.0",
               "TargetCPU": [
                 "x86",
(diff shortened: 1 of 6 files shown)


tst/CTA.Rules.Test/TempRules/microsoft.owin.security.notifications.json
```diff
@@ -20,16 +20,51 @@
           "Preferred": "Yes",
           "TargetFrameworks": [
             {
-              "Name": "net5.0",
+              "Name": "net6.0",
```
saikiranakula-amzn, line 23: why are we replacing target framework?

saikiranakula-amzn (approved):

tst/CTA.Rules.Test/TempRules/microsoft.owin.security.notifications.json
jonlouie: This field signifies that we should apply these actions if the target framework is `net6.0`. Previously, this rule was applied to both .NET 5 and .NET 6, and it added the package version `Microsoft.AspNetCore.Authentication.OpenIdConnect 5.0.12`.

This change updates the rule to only add that package version if we're targeting .NET 5. If .NET 6 is targeted, we get the latest version of the package (latest version is `6.0.0` and is only compatible with .NET 6).

Eruanion (approved):

PR #437, opened 2021-12-02
Added basic asp:TextBox control support

comsemer:
## Description
* Control converters will now pick up asp:TextBox components
* Added support for main attributes/events used by asp:TextBox
* Added node template for single tag nodes
* Note that code-behind references to asp:TextBox components and other
advanced features are still not being handled, as is the case with the
other control converters

## Supplemental testing
* Added unit tests for new control converter

---

### DIFF
--- a/tst/CTA.WebForms2Blazor.Tests/ControlConverters/TextBoxControlConverterTests.cs
+++ b/tst/CTA.WebForms2Blazor.Tests/ControlConverters/TextBoxControlConverterTests.cs
@@ -1,0 +1,82 @@
+﻿using CTA.WebForms2Blazor.ControlConverters;
+using HtmlAgilityPack;
+using NUnit.Framework;
+
+namespace CTA.WebForms2Blazor.Tests.ControlConverters
+{
+    public class TextBoxControlConverterTests
+    {
+        private TextBoxControlConverter _testConverter;
+
+        [SetUp]
+        public void SetUp()
+        {
+            _testConverter = new TextBoxControlConverter();
+        }
+
+        [Test]
+        public void Convert2Blazor_Returns_Standard_Input_For_SingleLine_Text_Mode()
+        {
+            var htmlString = @"<asp:TextBox ID=""TestId"" Text=""Example Text"">";
+            var expectedString = @"<input id=""TestId"" value=""Example Text"">";
+
+            var htmlNode = HtmlNode.CreateNode(htmlString, ControlConverter.PreserveCapitalization);
+            var convertedNode = _testConverter.Convert2Blazor(htmlNode);
+            var actualString = convertedNode.WriteTo();
+
+            Assert.AreEqual(expectedString, actualString);
+        }
+
+        [Test]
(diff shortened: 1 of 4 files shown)


saikiranakula-amzn (approved):

tst/CTA.WebForms2Blazor.Tests/ControlConverters/TextBoxControlConverterTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            var actualString = convertedNode.WriteTo();
+
+            Assert.AreEqual(expectedString, actualString);
+        }
+
+        [Test]
+        public void Convert2Blazor_Sets_ReadOnly_Attribute_Properly()
+        {
+            var htmlString = @"<asp:TextBox ID=""TestId"" Text=""Example Text"" ReadOnly=""true"">";
+            var expectedString = @"<input id=""TestId"" value=""Example Text"" readonly="""">";
```
jonlouie, line 60: Does the element created in `expectedString` function as expected? My understanding is that the `readonly` attribute does not take an assignment, for example: 
`<input id=""TestId"" value=""Example Text"" readonly>`

We may need to create a new `ViewLayerControlAttribute` to handle this case.

tst/CTA.WebForms2Blazor.Tests/ControlConverters/TextBoxControlConverterTests.cs
```diff
@@ -0,0 +1,82 @@
...
+            var actualString = convertedNode.WriteTo();
+
+            Assert.AreEqual(expectedString, actualString);
+        }
+
+        [Test]
+        public void Convert2Blazor_Sets_Disabled_Attribute_Properly()
+        {
+            var htmlString = @"<asp:TextBox ID=""TestId"" Text=""Example Text"" Enabled=""false"">";
+            var expectedString = @"<input id=""TestId"" value=""Example Text"" disabled="""">";
```
jonlouie, line 73: Same as above, but for the `disabled` attribute.

jonlouie (changes requested):

tst/CTA.WebForms2Blazor.Tests/ControlConverters/TextBoxControlConverterTests.cs
comsemer: I believe this format is also valid for boolean attributes, according to the html standard (https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes) the `readonly` attribute can be written as `readonly`, `readonly=readonly`, or `readonly=""`. If you want I can still make the change to handling boolean attributes the first way for stylistic reasons

jonlouie (approved):

tst/CTA.WebForms2Blazor.Tests/ControlConverters/TextBoxControlConverterTests.cs
jonlouie: Awesome, sounds good. Should be fine as-is, thanks!

PR #438, opened 2021-12-02
feat: WebForms CTA integration

kanikaul-amazon:
WebForms Project Integration With CTA

## Description
* Add WebForms handling in CTA Code Replacer
* Update W2B parameters to serve as a library for CTA

---


tst/CTA.Rules.Test/WebFormsTests.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        [TestCase(TargetFramework.Dotnet6)]
+        [TestCase(TargetFramework.Dotnet5)]
+        [TestCase(TargetFramework.DotnetCoreApp31)]
+        public void TestAspNetWebFormsProject(string version)
+        {
+            var solutionPath = CopySolutionFolderToTemp("ASP.NET-WebForms.sln", tempDir);
+            TestSolutionAnalysis results = AnalyzeSolution(solutionPath, version);
+
+            var project = results.ProjectResults.Where(prop => prop.CsProjectPath.EndsWith("ASP.NET-WebForms" + ".csproj")).FirstOrDefault();
```
marknfawaz, line 32: nit: Any reason why these strings are separated?

tst/CTA.Rules.Test/Models/ExpectedOutputConstants.cs
```diff
@@ -1647,5 +1647,81 @@ public static void Main(string[] args)
...
+            RouteConfig.RegisterRoutes(RouteTable.Routes);
+            BundleConfig.RegisterBundles(BundleTable.Bundles);
+        }
+
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddRazorPages();
+            services.AddServerSideBlazor();
+             // Did not attempt to migrate service layer
+            // and configure depenency injection in ConfigureServices(),
```
marknfawaz, line 1721: Fix spelling error: dependency*

marknfawaz (changes requested):
Before digging into the code, i have a high level question/concern. The CTA.Rules.Update project should not contain any porting logic, and we now have two branches in there (WCF & WebForms). Why can't these be in the PortCore project? Let me know if we need to discuss this.

jonlouie:
> Before digging into the code, i have a high level question/concern. The CTA.Rules.Update project should not contain any porting logic, and we now have two branches in there (WCF & WebForms). Why can't these be in the PortCore project? Let me know if we need to discuss this.

@marknfawaz I see what you mean. I still think it makes sense to modify `CTA.Rules.Update`, but a more sustainable approach might be to create `WcfProjectRewriter` and `WebFormsRewriter` classes, and use SolutionPort to call a factory to construct the appropriate project rewriter based on project type.

marknfawaz:
> > Before digging into the code, i have a high level question/concern. The CTA.Rules.Update project should not contain any porting logic, and we now have two branches in there (WCF & WebForms). Why can't these be in the PortCore project? Let me know if we need to discuss this.
> 
> @marknfawaz I see what you mean. I still think it makes sense to modify `CTA.Rules.Update`, but a more sustainable approach might be to create `WcfProjectRewriter` and `WebFormsRewriter` classes, and use SolutionPort to call a factory to construct the appropriate project rewriter based on project type.

@jonlouie "I still think it makes sense to modify CTA.Rules.Update" Not sure what u mean by that. As mentioned, the update project shouldn't be aware of how the actions are used, it should just run these actions. As for the Rewriters, yes, that should be our approach, but it should be similar to what we have in the PortCore project.

src/CTA.Rules.PortCore/ProjectSpecificPort/ProjectRewriters/WebFormsProjectRewriter.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            return _projectResult;
+        }
+
+        public override List<IDEFileActions> RunIncremental(List<string> updatedFiles, RootNodes projectRules)
+        {
+            var ideFileActions = new List<IDEFileActions>();
+
+            // Incremental porting for WebForms project is yet to be implemented
```
marknfawaz, line 69: throw a NotImplementedException or log a call to this so we can figure out if this was used

src/CTA.Rules.PortCore/ProjectSpecificPort/WCF/Constants.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System.Collections.Generic;
 
-namespace CTA.Rules.Update.WCF
+namespace CTA.Rules.PortCore.WCF
```
marknfawaz, line 3: Why is this modified while the file is still in the Update project?

src/CTA.Rules.PortCore/ProjectSpecificPort/WCF/WCFServicePort.cs
```diff
@@ -14,7 +14,7 @@
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using Constants = CTA.Rules.Update.WCF.Constants;
+using Constants = CTA.Rules.PortCore.WCF.Constants;
 
 namespace CTA.Rules.PortCore
 {
```
marknfawaz, line 20: Does this file belong in this project?

marknfawaz (changes requested):

src/CTA.Rules.PortCore/ProjectSpecificPort/WCF/WCFServicePort.cs
jonlouie: This was moved to PortCore project and is only referenced in the WCFProjectRewriter which also lives in this project. Same as above with Github trying to display this in the diff view

src/CTA.Rules.PortCore/ProjectSpecificPort/WCF/Constants.cs
jonlouie: This file was moved to the PortCore project. Github's diff view tries to show that, but it gets cut off depending on browser window size

src/CTA.Rules.PortCore/ProjectSpecificPort/ProjectRewriters/WebFormsProjectRewriter.cs
jonlouie: Will add a log statement

PR #441, opened 2021-12-06
Improve error handling

jonlouie:
## Related issue

Closes: #ISSUE_NUMBER_HERE

## Description
* Improves error handling in the following files:
  * Migration Manager 
  * ClassConverterFactory
  * FileConverterFactory
  * ViewFileConverter 
  * CodeFileConverter
  * ControlConverter

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
+++ b/src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
@@ -1,9 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
-using Codelyzer.Analysis.CSharp;
+using CTA.Rules.Config;
 using CTA.WebForms2Blazor.ClassConverters;
 using CTA.WebForms2Blazor.Factories;
 using CTA.WebForms2Blazor.FileInformationModel;
@@ -19,7 +18,7 @@ namespace CTA.WebForms2Blazor.FileConverters
         private readonly WorkspaceManagerService _blazorWorkspaceBuilder;
         private readonly ProjectAnalyzer _webFormsProjectAnaylzer;
         private readonly ClassConverterFactory _classConverterFactory;
-        private readonly IEnumerable<ClassConverter> _classConverters;
+        private readonly IEnumerable<ClassConverter> _classConverters = new List<ClassConverter>();
 
         public CodeFileConverter(
             string sourceProjectPath,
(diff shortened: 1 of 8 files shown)


src/CTA.WebForms2Blazor/ProjectManagement/ProjectBuilder.cs
```diff
@@ -33,21 +41,28 @@ public void WriteFileInformationToProject(FileInformation newDocument)
...
+                    stream.Write(fileContent, 0, fileContent.Length);
+                }
 
-            using (FileStream stream = File.Create(fullPath))
+                LogHelper.LogInformation(string.Format(FileInfoWrittenLogTemplate, GetType().Name, fullPath));
+            }
+            catch (Exception e)
             {
-                stream.Write(fileContent, 0, fileContent.Length);
+                LogHelper.LogError(e, "Could not write file bytes to project.");
```
comsemer, line 64: May want to include the file location we attempted to write to here

src/CTA.WebForms2Blazor/ProjectManagement/ProjectBuilder.cs
```diff
@@ -20,10 +21,17 @@ public ProjectBuilder(string outputProjectPath)
...
+                var fullPath = Path.Combine(_outputProjectPath, relativePath);
 
-            // No exception is thrown if the directory already exists
-            Directory.CreateDirectory(fullPath);
+                // No exception is thrown if the directory already exists
+                Directory.CreateDirectory(fullPath);
+            }
+            catch (Exception e)
+            {
+                LogHelper.LogError(e, "Could not create relative directory.");
```
comsemer, line 33: May want to record the directory location in the error message here

src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
```diff
@@ -42,8 +41,16 @@ public class CodeFileConverter : FileConverter
 
             // This code is set up to not break unit tests but still work for the demo
             // use code above after demo and just fix unit tests
-            _fileModel = _webFormsProjectAnaylzer.AnalyzerResult.ProjectBuildResult?.SourceFileBuildResults?
-                .Single(r => r.SourceFilePath.EndsWith(RelativePath))?.SemanticModel;
+            try
+            {
+                _fileModel = _webFormsProjectAnaylzer.AnalyzerResult.ProjectBuildResult?.SourceFileBuildResults?
+                    .Single(r => r.SourceFileFullPath.EndsWith(RelativePath))?.SemanticModel;
```
comsemer, line 47: For this line, you'll want to remove the null checks to get the exception to be thrown. `...ProjectBuildResult?.SourceFileBuildResults?
                    .Single(r => r.SourceFileFullPath.EndsWith(RelativePath))?.SemanticModel;` should become `...ProjectBuildResult.SourceFileBuildResults
                    .Single(r => r.SourceFileFullPath.EndsWith(RelativePath)).SemanticModel;` otherwise `_fileModel` will default to null, but the error message won't be recorded

comsemer (changes requested):
Looks pretty good, really only 1 substantive change requested, the other 2 are just nitpicks about error message content

src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
jonlouie: I left the null checks because we could catch an error on `.Single()` which will throw an error if it doesn't return exactly 1 result (ran into that issue during some test runs).

I don't think returning a null value the exceptional behavior here, but rather if an error is caught, we just default to null and let the user know. If you disagree, we can remove null checks so we can log those cases.

src/CTA.WebForms2Blazor/ProjectManagement/ProjectBuilder.cs
jonlouie: The path might be recorded by the exception itself, but will add here explicitly just in case

src/CTA.WebForms2Blazor/ProjectManagement/ProjectBuilder.cs
jonlouie: The path might be recorded by the exception itself, but will add here explicitly just in case

src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
comsemer: Sounds good, we can leave this as is

comsemer (approved):

PR #442, opened 2021-12-06
Code conversion improved error handling

comsemer:
## Description
Added extra error handling for:
* GlobalClassConverter
* HttpHandlerClassConverter
* HttpModuleClassConverter
* PageCodeBehindClassConverter
* CodeFileConverter

---

### DIFF
--- a/src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
+++ b/src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
@@ -4,6 +4,7 @@ using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Codelyzer.Analysis.CSharp;
+using CTA.Rules.Config;
 using CTA.WebForms2Blazor.ClassConverters;
 using CTA.WebForms2Blazor.Factories;
 using CTA.WebForms2Blazor.FileInformationModel;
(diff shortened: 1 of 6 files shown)


jonlouie (approved):

src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
```diff
@@ -58,19 +50,44 @@ public class CodeFileConverter : FileConverter
...
+            var allMigrationTasks = Task.WhenAll(classMigrationTasks.Select(t => t.Item2));
+
+            try
+            {
+                await allMigrationTasks;
+            }
+            // NOTE: We don't provide an reference to the caught exception because it would just be a
+            // standard aggregate exception of the exceptions that arose in the migration tasks, instead
+            // we choose to log an exception for each of the failed tasks to provide more info
+            catch
```
saikiranakula-amzn, line 79: shouldn't we catch    catch (AggregateException ex){}

saikiranakula-amzn (approved):

src/CTA.WebForms2Blazor/FileConverters/CodeFileConverter.cs
comsemer: I don't think we would benefit from the reference to the aggregate exception as in the catch block we are enumerating then logging the exceptions that would make up the aggregate anyway. I think this is preferable to logging the aggregate exception because it will make our metrics on number and type of exceptions more accurate

PR #443, opened 2021-12-07
Metrics web forms action

saikiranakula-amzn:
## Related issue

Closes: #ISSUE_NUMBER_HERE

## Description

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.WebForms2Blazor/ClassConverters/ControlCodeBehindClassConverter.cs
+++ b/src/CTA.WebForms2Blazor/ClassConverters/ControlCodeBehindClassConverter.cs
@@ -6,6 +6,7 @@ using System.Threading.Tasks;
 using CTA.WebForms2Blazor.Extensions;
 using CTA.WebForms2Blazor.FileInformationModel;
 using CTA.WebForms2Blazor.Helpers;
+using CTA.WebForms2Blazor.Metrics;
 using CTA.WebForms2Blazor.Services;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -14,22 +15,26 @@ namespace CTA.WebForms2Blazor.ClassConverters
 {
     public class ControlCodeBehindClassConverter : ClassConverter
     {
+        private const string ActionName = "ControlCodeBehindClassConverter";
+        private WebFormMetricContext _metricsContext;
         public ControlCodeBehindClassConverter(
             string relativePath,
             string sourceProjectPath,
             SemanticModel sourceFileSemanticModel,
             TypeDeclarationSyntax originalDeclarationSyntax,
             INamedTypeSymbol originalClassSymbol,
-            TaskManagerService taskManager)
+            TaskManagerService taskManager,
+            WebFormMetricContext metricsContext)
             : base(relativePath, sourceProjectPath, sourceFileSemanticModel, originalDeclarationSyntax, originalClassSymbol, taskManager)
         {
+            _metricsContext = metricsContext;
             // TODO: Register with the necessary services
         }
 
         public override Task<IEnumerable<FileInformation>> MigrateClassAsync()
         {
             LogStart();
-
+            _metricsContext.CollectClassConversionMetrics(ActionName);
             // NOTE: Removed temporarily until usings can be better determined, at the moment, too
             // many are being removed
             //var requiredNamespaces = _sourceFileSemanticModel.GetNamespacesReferencedByType(_originalDeclarationSyntax);
--- a/src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
+++ b/src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
@@ -20,7 +20,7 @@ namespace CTA.WebForms2Blazor.FileConverters
     {
         private const string ChildActionType = "ViewFileConverter";
         private const string UnSupportedControlConverter = "UnSupportedControlConverter";
-        private Regex AspControlsRegex = new Regex(@"<asp:(\w+)\s+.*?id=(\w +)");
+        private Regex AspControlsRegex = new Regex(@"\w+:\w+");
         private ViewImportService _viewImportService;
         private List<ControlConversionAction> _controlActions;
         private readonly WebFormMetricContext _metricsContext;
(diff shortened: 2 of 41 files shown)


src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -66,23 +73,25 @@ private void FindConversionActions(HtmlNode node, HtmlNode parent)
...
             else if (SupportedControls.UserControls.UserControlRulesMap.ContainsKey(node.Name))
             {
-                var conversionAction = new ControlConversionAction(node, parent,
-                    SupportedControls.UserControls.UserControlRulesMap[node.Name]);
+                var conversionAction = new ControlConversionAction(node, parent, SupportedControls.UserControls.UserControlRulesMap[node.Name]);
+                controlConverterType = conversionAction.ControlConverter.GetType().Name;
                 _controlActions.Add(conversionAction);
             }
-            else
+            else if(UnknownControlRemover.ControlStartTagRegex.IsMatch(node.Name) || UnknownControlRemover.ControlStartTagRegex.IsMatch(node.Name))
```
comsemer, line 89: The regexes `ControlStartTagRegex` and `ControlEndTagRegex` matches the entire tag including the `<>` and `</>` brackets along with any attributes whereas `node.Name` does not and only includes the name of the tag. For this reason, this condition should never be true. Also, you are using `ControlStartTagRegex` on both sides of the `||`.

src/CTA.WebForms2Blazor/ClassConverters/ControlCodeBehindClassConverter.cs
```diff
@@ -14,22 +15,26 @@ namespace CTA.WebForms2Blazor.ClassConverters
 {
     public class ControlCodeBehindClassConverter : ClassConverter
     {
+        private const string ActionName = "ControlCodeBehindClassConverter";
```
jonlouie, line 18: We can improve maintenance and repetition by adding `ActionName` as a property to the base class `ClassConverter` and defining it as:

```csharp
public class ClassConverter
{
    protected virtual string ActionName => GetType().Name;
    ...
}
```

This can be done for all converter base classes

src/CTA.WebForms2Blazor/Helpers/ControlHelpers/EmbeddedCodeReplacers.cs
```diff
@@ -116,11 +117,12 @@ public static string ReplaceEmbeddedCodeBlocks(string htmlString)
                 match.Groups[EmbeddedExpressionRegexGroupName].Value.Trim()));
         }
 
-        public static string ConstructBlazorDirectives(string content, string originalFilePath, string projectName, ViewImportService viewImportService)
+        public static string ConstructBlazorDirectives(string content, string originalFilePath, string projectName, ViewImportService viewImportService, WebFormMetricContext metricContext)
         {
             var directiveName = DirectiveNameRegex.Match(content).Groups[DirectiveNameRegexGroupName].Value;
             var directiveConverter = SupportedControls.DirectiveRulesMap.ContainsKey(directiveName) ?
                 SupportedControls.DirectiveRulesMap[directiveName] : SupportedControls.DefaultDirectiveConverter;
+            metricContext.CollectDirectiveConversionMetrics(directiveName+"DirectiveConvertor");
```
comsemer, line 125: Typo: "DirectiveConverter"

src/CTA.WebForms2Blazor/ClassConverters/ControlCodeBehindClassConverter.cs
saikiranakula-amzn: I was thinking that as well, only reason I used the constants is if someone don't name the class name properly. I will make the change

comsemer (changes requested):
Looks good for the most part, just one issue with the condition checking unsupported controls and a typo

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -3,31 +3,39 @@
...
 using CTA.WebForms2Blazor.Services;
 using HtmlAgilityPack;
 
 namespace CTA.WebForms2Blazor.FileConverters
 {
     public class ViewFileConverter : FileConverter
     {
+        private const string ChildActionType = "ViewFileConverter";
+        private const string UnSupportedControlConverter = "UnSupportedControlConverter";
+        private Regex AspControlsRegex = new Regex(@"<asp:(\w+)\s+.*?id=(\w +)");
```
comsemer, line 23: Still a few slight issues with this regex, for an unsupported control `node.Name` below would look something like this: _asp:unsupportedcontrol_ but it might also look like _yyy:othercontrol_. To match this, a simple expression like `\w+:\w+` would do fine. I'm not sure if the `?id=(\w+)` is meant to be a capture group but if so then the expression could be modified to look like this `\w+:(?<id>\w+)`

src/CTA.WebForms2Blazor/FileConverters/ViewFileConverter.cs
```diff
@@ -66,23 +74,30 @@ private void FindConversionActions(HtmlNode node, HtmlNode parent)
...
                 _controlActions.Add(conversionAction);
             }
             else
             {
-                // No conversion found found for this node type
-                // TODO: add metrics here if node.Name matches pattern: [A-z]*:[A-z]*
-                LogHelper.LogInformation($"No porting action found for node type: {node.Name}");
+                Match aspControlTagRegex = AspControlsRegex.Match(node.Name);
+                Match matchStartTagRegex = UnknownControlRemover.ControlStartTagRegex.Match(node.Name);
+                Match matchEndTagRegex = UnknownControlRemover.ControlEndTagRegex.Match(node.Name);
```
comsemer, line 94: In a similar vein to the comment above, the custom regex should be good enough for this check, `matchStartTagRegex` and `matchEndTagRegex` should never successfully match `node.Name`. Just for reference, the reason these regexes work later on when unsupported controls are removed is because they are compared against the entire view file content rather than a single node name

comsemer (commented):
Still a couple of fixes to make with the unknown control match regex

comsemer (approved):

PR #449, opened 2021-12-09
W2b kanikaul edit

saikiranakula-amzn:
## Related issue

Closes: #ISSUE_NUMBER_HERE

## Description

## Supplemental testing

## Additional context

---


src/CTA.Rules.Metrics/MetricsTransformer.cs
```diff
@@ -121,6 +122,25 @@ internal static IEnumerable<MissingMetaReferenceMetric> TransformMissingMetaRefe
...
+            foreach (var metric in projectResult.WebFormsMetricResults)
+            {
+                if(metric.ActionName == WebFormsActionType.FileConversion)
+                    webFormActionMetrics.Add(new FileConversionMetric(context, metric.ChildAction, projectFile));
+                else if(metric.ActionName == WebFormsActionType.ControlConversion)
+                    webFormActionMetrics.Add(new ControlConversionMetric(context,metric.ChildAction,metric.NodeName, projectFile));
+                else if(metric.ActionName == WebFormsActionType.ClassConversion)
+                    webFormActionMetrics.Add(new ClassConversionMetric(context, metric.ChildAction, projectFile));
+                else if(metric.ActionName == WebFormsActionType.DirectiveConversion)
+                    webFormActionMetrics.Add(new DirectiveConversionMetric(context, metric.ChildAction, projectFile));
```
jonlouie, line 138: We should log a warning or error if ActionName does not match a WebFormsActionType

src/CTA.Rules.Metrics/PortSolutionResultReportGenerator.cs
```diff
@@ -103,13 +106,19 @@ private void GenerateMetrics()
                 actionPackageMetrics.AddRange(MetricsTransformer.TransformActionPackages(Context, projectResult));
                 actionExecutionMetrics.AddRange(MetricsTransformer.TransformGenericActionExecutions(Context, projectResult));
                 missingMetaReferenceMetrics.AddRange(MetricsTransformer.TransformMissingMetaReferences(Context, projectResult));
+
+                if (projectResult.WebFormsMetricResults != null)
+                {
+                    webFormsActionMetrics.AddRange(MetricsTransformer.TransformWebFormActionMetrics(Context, projectResult));
+                }
```
jonlouie, line 113: Rather than check for a null here, it may be cleaner and safer to always initialize `WebFormsMetricResults` to an empty collection in the `ProjectResult` constructor

src/CTA.Rules.Metrics/PortSolutionResultReportGenerator.cs
```diff
@@ -140,6 +149,11 @@ private string GeneratePortSolutionResultJsonReport()
                 .Concat(MissingMetaReferenceMetrics)
                 .ToList();
 
+            if (WebFormsActionMetrics != null)
+            {
+                allMetrics = allMetrics.Concat(WebFormsActionMetrics).ToList();
+            }
```
jonlouie, line 155: Similar to above, if we always initialize this to an empty collection, we can remove the if statement and always concat it.

jonlouie (commented):
Couple of minor comments/changes, otherwise looks ok. We'll also need to add/update some tests for the report generator, but we can add that in another PR.

One larger concern is that our metrics integration further couples CTA and W2B, making a spinoff more difficult. In the short term, this will work, but longer term we will probably have to do a refactor and use dependency injection to pass in a metrics collection service.

PR #461, opened 2021-12-14
Add Error tag for categorization

kanikaul-amazon:
## Description
* Add Error Tag to WebForms porting.
* This will be used to classify errors in metrics

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Config/Constants.cs
+++ b/src/CTA.Rules.Config/Constants.cs
@@ -148,5 +148,8 @@ namespace CTA.Rules.Config
             new List<string> { "wcfconfigbasedservice", "Program.cs"},
             new List<string> { "wcfconfigbasedservice", "Startup.cs"}
         };
+
+        public const string WCFErrorTag = "WCF Porting Error: ";
+        public const string WebFormsErrorTag = "WebForms Porting Error: ";
     }
 }
--- a/src/CTA.WebForms2Blazor/ClassConverters/GlobalClassConverter.cs
+++ b/src/CTA.WebForms2Blazor/ClassConverters/GlobalClassConverter.cs
@@ -99,7 +99,7 @@ namespace CTA.WebForms2Blazor.ClassConverters
             }
             catch (Exception e)
             {
-                LogHelper.LogError(e, $"{Constants.WebFormsErrorTag}Failed to construct new Startup file content from {OriginalClassName} class at {_fullPath}");
+                LogHelper.LogError(e, $"{Rules.Config.Constants.WebFormsErrorTag}Failed to construct new Startup file content from {OriginalClassName} class at {_fullPath}");
             }
 
             DoCleanUp();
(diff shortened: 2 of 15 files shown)


src/CTA.WebForms2Blazor/ClassConverters/GlobalClassConverter.cs
```diff
@@ -171,7 +171,7 @@ private async Task<IEnumerable<UsingDirectiveSyntax>> GetAllUsingStatements()
             }
             catch (OperationCanceledException e)
             {
-                LogHelper.LogError(e, string.Format(
+                LogHelper.LogError(e, Constants.WebFormsErrorTag, string.Format(
                     Constants.CaneledServiceCallLogTemplate,
                     GetType().Name,
                     typeof(LifecycleManagerService).Name,
```
jonlouie, line 177: We might need to refactor this- for this particular signature of `LogError()`, the 2nd parameter is expected to be a string template and all following params are substituted into that string template. It might be safer to prepend the tag into the log template.

src/CTA.WebForms2Blazor/ClassConverters/GlobalClassConverter.cs
```diff
@@ -216,7 +216,8 @@ private async Task InsertRequestPipelineMiddlewareRegistrations()
             }
             catch (OperationCanceledException e)
             {
-                LogHelper.LogError(e, string.Format(
+                LogHelper.LogError(e, Constants.WebFormsErrorTag, 
+                    string.Format(
```
jonlouie, line 220: Same as above

src/CTA.Rules.Config/Constants.cs
```diff
@@ -148,5 +148,8 @@ public class Constants
             new List<string> { "wcfconfigbasedservice", "Program.cs"},
             new List<string> { "wcfconfigbasedservice", "Startup.cs"}
         };
+
+        public const string WCFErrorTag = "WCF Porting Error: ";
+        public const string WebFormsErrorTag = "WebForms Porting Error: ";
```
jonlouie, line 153: Do we need `WebFormsErrorTag ` defined in two places? It is also defined in `src/CTA.WebForms2Blazor/Constants.cs`. If it does not create a circular dependency, I recommend leaving it defined in W2B and removing it from here.

src/CTA.WebForms2Blazor/MigrationManager.cs
```diff
@@ -93,7 +93,9 @@ public async Task<WebFormsPortingResult> PerformMigration()
                     }
                     catch (Exception e)
                     {
-                        LogHelper.LogError(e, string.Format(Constants.OperationFailedLogTemplate, GetType().Name, Constants.FileMigrationLogAction));
+                        LogHelper.LogError(
+                            e, string.Join("", Constants.WebFormsErrorTag,
+                        string.Format(Constants.OperationFailedLogTemplate, GetType().Name, Constants.FileMigrationLogAction)));
```
jonlouie, line 98: Do we need the empty string in "string.Join()" for anything?

jonlouie (approved):

Issue #462, opened 2021-12-14
Web Forms maintenance 


jonlouie:
## Description of issue
Based on comments reported in https://github.com/aws/cta/pull/436:
* Remove .DS_Store files
* Verify exclusion of .DS_store files .gitignore
* Verify INamespaceSymbol equality comparer
* Move SolutionHash and ProjectGUID properties to a base class in Web Forms metrics
* Remove reference to w2b from CTA.Rules.Update.csproj

PR #466, opened 2021-12-15
W2b cleanup

jonlouie:
## Related issue

Closes: #462

## Description
* Remove .DS_Store files
* Verify exclusion of .DS_store files .gitignore
* Replace INamedSymbol with namespace strings to ensure proper comparison
* Move SolutionHash and ProjectGUID properties to a base class in Web Forms metrics
* Move reference to w2b in CTA.Rules.Update.csproj to CTA.Rules.PortCore.csproj

---


saikiranakula-amzn (dismissed):

kanikaul-amazon (approved):

tst/CTA.WebForms2Blazor.Tests/Extensions/SemanticModelExtensionTests.cs
```diff
@@ -117,7 +117,7 @@ public async Task GetNamespacesReferencedByType_Does_Not_Contain_Own_Namespace()
             var classSymbol = model.GetDeclaredSymbol(testClass1Declaration);
             var requiredNamespaces = model.GetNamespacesReferencedByType(testClass1Declaration);
 
-            Assert.False(requiredNamespaces.Contains(classSymbol?.ContainingNamespace));
+            Assert.False(requiredNamespaces.Contains(classSymbol?.ContainingNamespace.ToDisplayString()));
```
marknfawaz, line 120: Can this be null?

marknfawaz (changes requested):
Please make sure DS_Store files are removed

tst/CTA.WebForms2Blazor.Tests/Extensions/SemanticModelExtensionTests.cs
comsemer: I don't think we want to allow `classSymbol.ContainingNamespace` to be null here, as we want to assume that any class symbol we are processing belongs to a namespace, so we can probably get rid of the null conditional operator here, though I don't think `classSymbol.ContainingNamespace` should evaluate to null in the first place

comsemer (commented):
Looks good to me, may just want to remove the null conditional operator near Mark's comment

tst/CTA.WebForms2Blazor.Tests/Extensions/SemanticModelExtensionTests.cs
jonlouie: According to docs, ContainingNamespace can be null. It's possible to define classes without a namespace block. I believe in those cases, ContainingNamespace would be null.

comsemer: Ah ok, makes sense, thanks for clearing that up

comsemer (approved):
Looks good to me

marknfawaz (approved):

Issue #473, opened 2021-12-17
Read project paths from solution file rather than finding them through directories


Eruanion:
## Description
We should read the project paths from the solution file rather than attempt to discover them through the directory structure of where the solution file is.

## Solution idea(s)
Read from the solution file.

PR #474, opened 2021-12-17
Read project paths from solution file

Eruanion:
## Related issue

Closes: #473 

## Description
We should read the project paths from the solution file rather than attempt to discover them through the directory structure of where the solution file is. This method is now stored in codelyzer repo. This PR ensures we use this everywhere in CTA.

This also updated all our projects to .net 6 as a requirement as it consumes latest codelyzer which is also .net 6 based.

## Supplemental testing
All unit tests passed.

## Additional context

---


src/CTA.Rules.Config/Utils.cs
```diff
@@ -15,6 +16,13 @@ namespace CTA.Rules.Config
     public class Utils
     {
         private const string DefaultMutexName = "DefaultMutex";
+        private static List<SolutionProjectType> AcceptedProjectTypes = new List<SolutionProjectType>()
```
Eruanion, line 19: Please pay special attention to this list of allowed project types, the entire list is:
SolutionProjectType.SharedProject,
SolutionProjectType.SolutionFolder,
SolutionProjectType.Unknown,
The first two of which are non-buildable. Unknown is most likely not buildable.

marknfawaz: Why do we need a list of these types, and what happens if a project isn't of that type? Do we ignore while copying/porting?

marknfawaz (changes requested):

marknfawaz:
Couple questions:
1. What happens if we receive a Project/PortCore Configuration object with a different set of csprojs?
2. Did you check if codelyzer does the same thing now (or do we need to change it)

I'm trying to think of the implications on solutions using/used by CTA

src/CTA.Rules.Config/Utils.cs
Eruanion: So without this list of types, many non-project like references get picked up within the solution file as "projects", one of the worse perpetrators of this is a folder. If there is a folder in the root solution file it gets defined as a "project" in that solution file. Which means we will attempt to build it and port it and this causes unexpected errors. Before these would just be ignored as we were scanning folder structure for .csproj type files only. So with this new code we ignore any non build-able type of project essentially. I am not sure about porting as it is still part of the solution, it just doesn't get defined as a project specifically, so I would still expect it to get copied but not to be treated as a project to run rules against.

Eruanion:
> Couple questions:
> 
>     1. What happens if we receive a Project/PortCore Configuration object with a different set of csprojs?
> 
>     2. Did you check if codelyzer does the same thing now (or do we need to change it)
> 
> 
> I'm trying to think of the implications on solutions using/used by CTA

1. For this question we do have an issue to rectify this in PA client as well so we also parse the solution file, so eventually these should always end up with same project references but until that ticket is completed CTA will still port as if it had found the projects through the old method with folder scanning for .csproj projects and behave the same as before from there on.

2. Currently codelyzer already parses the solution file exactly as it is being done with this new code. The only difference is that it doesn't check for project types (could be an issue in the future) and there is not error catching surrounding it in case the solution file is non parse-able. In the new CTA code included here we fall back to folder scanning for .csproj files the way we use to do it before if the solution file fails to parse for some reason.

tst/CTA.Rules.Test/AwsRulesBaseTest.cs
```diff
@@ -139,22 +143,38 @@ public string GetSrcPath(string path)
...
+                        projectResults.AppendLine(projectResult.ProjectActions.ToString());
+                        result.ProjectResults.Where(p => p.CsProjectPath == projectResult.ProjectFile).FirstOrDefault().ProjectAnalysisResult = projectResults.ToString();
+
+                        str.Append(projectResults);
                     }
+                    result.SolutionAnalysisResult = str.ToString();
+
+
                     var runResult = solutionPort.Run();
-                    result = GenerateSolutionResult(solutionDir, analysisRunResult, runResult);
```
Beau-Gosse-dev, line 147: Could you help me understand why we changed how we build the result. I'm not quite seeing how reading project paths from the solution file affects all this.

Eruanion: Oh this was a side fix. I realized that we were iterating over the entire project list twice and re-doing a lot of the same work. So I remove this reference and decided to take advantage of current loops to fill in the return object instead, which sped up the UTs a bit also, but more importantly is also much more efficient.

src/CTA.Rules.Config/Utils.cs
```diff
@@ -267,5 +275,27 @@ public static void DownloadFilesToFolder(string s3Bucket, string targetFolder, L
...
+                {
+                    LogHelper.LogError(ex, $"Error while parsing solution file {solutionPath} falling back to directory parsing.");
+                    projectPaths = Directory.EnumerateFiles(solutionDir, "*.csproj", SearchOption.AllDirectories);
+                }
+
+                return projectPaths;
+            }
+            return null;
+        }
     }
```
marknfawaz, line 300: what are your thoughts on moving this to codelyzer so we can use in all the projects using it?

Eruanion: Agreed, moving to codelyzer, removing project types to consume all but move pruning logic to build function to only attempt to build projects of certain types.

src/CTA.Rules.Config/Utils.cs
```diff
@@ -267,5 +275,27 @@ public static void DownloadFilesToFolder(string s3Bucket, string targetFolder, L
...
+                }
+                catch(Exception ex)
+                {
+                    LogHelper.LogError(ex, $"Error while parsing solution file {solutionPath} falling back to directory parsing.");
+                    projectPaths = Directory.EnumerateFiles(solutionDir, "*.csproj", SearchOption.AllDirectories);
+                }
+
+                return projectPaths;
+            }
+            return null;
```
jonlouie, line 298: Can we return an empty collection instead of null? This method replaces a `Directory.EnumerateFiles()` call which returns an empty collection when nothing is found.

src/CTA.Rules.Config/Utils.cs
jonlouie, line 19: nit: Can we change `AcceptedProjectTypes` to a HashSet?

jonlouie (changes requested):

src/CTA.Rules.Config/Utils.cs
Eruanion: absolutely agreed

src/CTA.Rules.Config/Utils.cs
Eruanion: also can do yup

jonlouie (commented):
LGTM but will hold off on approval until PR https://github.com/aws/cta/pull/482 is merged first

saikiranakula-amzn (approved):

Beau-Gosse-dev (approved):

comsemer (approved):

jonlouie (approved):

marknfawaz (approved):

Issue #481, opened 2021-12-23
Upgrade to .NET 6


Beau-Gosse-dev:
## Description
Upgrade project to .NET 6 to keep it up to date with latest version

## Solution idea(s)
Update all project files

PR #482, opened 2021-12-23
Upgrade to .NET 6

Beau-Gosse-dev:
## Related issue

Closes: [#ISSUE_NUMBER_HERE ](https://github.com/aws/cta/issues/481)

## Description
* Upgrade all projects to .NET 6
* Add missing CLI argument to README
* Fixed build warnings
* Update all packages and use new Codelyzer v2
* This uses the new version of Codelyzer that doesn't pass a language version argument anymore

## Supplemental testing
I ran the command line locally against a webapi project and eShopOnBlazor. Both had errors, but they were the same errors I got on main branch too, so I don't think they were related to this change and most of the solution ported. 

I also referenced this from the PA Client CLI and analyzed OwinParadise which resulted in the same results as current main code. 

## Additional context
The README change was found when I went to test this locally. The package lock change was added because as I upgrade to .NET 6 a bunch of packages and lines in there changed and it got me wondering why we locked those anyway. The exception change and symbol comparison type changes were to fix .NET 6 build warnings. This was build and tested with VS2022.

---


jonlouie (changes requested):
Awesome, thanks Beau. We use the packages.lock.json to prevent remote code execution (the highlighted part of [this article](https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/#:~:text=Package%20content%20mismatch,new%20direct%20dependency.) explains it well) but otherwise this looks great.

saikiranakula-amzn (dismissed):

jonlouie (dismissed):

src/CTA.Rules.Config/Utils.cs
```diff
@@ -109,14 +109,14 @@ public static string DownloadFile(string fileUrl, string destinationFile, int re
...
-                    catch (Exception ex)
+                    catch (Exception)
                     {
                         retryAttempts++;
                         Thread.Sleep(Constants.DefaultThreadSleepTime);
 
                         if (retryAttempts == retryCount)
                         {
-                            throw ex;
+                            throw;
```
marknfawaz, line 119: Why remove? shouldn't we throw the original exception so that the function using it has data around why this happened?

marknfawaz (changes requested):
Thanks for updating. Can we update the versions of the packages used in the next commit (this includes codelyzer), or do u think that should be part of a different PR?

src/CTA.Rules.Config/Utils.cs
Beau-Gosse-dev: This will still throw the original exception. The point of this is to maintain the stack trace of the original exception.

I originally thought this had come up as a build warning, which is why I changed it. But looking back, it might have just been a code quality warning. https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2200

By just calling "throw" instead of "throw ex" we preserver the stack trace of the error. So we would be able to see exactly where the real error was thrown. If we call "throw ex" the stack trace will only point us to this line here, which I assume won't be as helpful.

marknfawaz (dismissed):

comsemer (approved):

saikiranakula-amzn (dismissed):

jonlouie (approved):
LGTM. Thx, Beau

comsemer (dismissed):

marknfawaz (dismissed):

jonlouie (dismissed):

comsemer (approved):

jonlouie (approved):

marknfawaz (approved):

saikiranakula-amzn (approved):

PR #510, opened 2022-01-18
AggregateException Removal

comsemer:
Removed logging of AggregateExceptions and ensured that potential
AggregateExceptions are unwrapped properly

Unit tests were unaffected by this change

---

### DIFF
--- a/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
+++ b/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
@@ -61,7 +61,9 @@ namespace CTA.WebForms.FileConverters
             LogStart();
             _metricsContext.CollectActionMetrics(WebFormsActionType.FileConversion, ChildActionType);
 
-            var classMigrationTasks = _classConverters.Select(converter => converter.MigrateClassAsync());
+            // Need to have ToList call here to enumerate the collection and ensure class
+            // converters are running before we retire this file converter task
+            var classMigrationTasks = _classConverters.Select(converter => converter.MigrateClassAsync()).ToList();
 
             // We want to do our cleanup now because from this point on all migration tasks
             // are done by class converters and we want to make sure that we retire the task
--- a/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
+++ b/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
@@ -80,17 +80,27 @@ namespace CTA.WebForms.FileConverters
             {
                 await allMigrationTasks;
             }
-            // NOTE: We use Exception here instead of AggregateException as sometimes await
-            // will auto un-wrap aggregate exception
-            catch (Exception e)
+            // NOTE: We don't provide a reference to the caught exception here because any
+            // encountered exceptions will be logged when examining the individual failed tasks
+            catch
             {
-                LogHelper.LogError(e, $"{Rules.Config.Constants.WebFormsErrorTag}Collection of migration tasks experienced 1 or more failures");
+                // NOTE: While using await does automatically unwrap AggregateExceptions when they
+                // only contain 1 exception, component tasks of Task.WhenAll() will never have their
+                // AggregateExceptions unwrapped when the WhenAll task is awaited
 
-                var failedMigrationTasks = classMigrationTasks.Where(t => t.Item2.Status != TaskStatus.RanToCompletion);
+                var failedMigrationTasks = classMigrationTasks
+                    // Select failed tasks only
+                    .Where(t => t.Item2.Status != TaskStatus.RanToCompletion)
+                    // We can be fairly certain that failed tasks will only have a single exception,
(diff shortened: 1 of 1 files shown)


src/CTA.WebForms/FileConverters/CodeFileConverter.cs
```diff
@@ -80,17 +80,27 @@ public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
             {
                 await allMigrationTasks;
             }
-            // NOTE: We use Exception here instead of AggregateException as sometimes await
-            // will auto un-wrap aggregate exception
-            catch (Exception e)
+            // NOTE: We don't provide a reference to the caught exception here because any
+            // encountered exceptions will be logged when examining the individual failed tasks
+            catch
```
saikiranakula-amzn, line 85: I think this approach works fine, but can we do this instead:
catch (AggregateException ex) {
ex.flatten().Handle(e => {"Log error message"});
}
just a suggestion to make it look simple

comsemer: While that is simple and clean, we would lose some context on how the exceptions are currently being logged below. The line that logs the errors
`LogHelper.LogError(failedTask.Item2, $"{Rules.Config.Constants.WebFormsErrorTag}Failed to migrate {failedTask.Item1.OriginalClassName} class " +
                        $"located at {failedTask.Item1.FullPath}");`
is generated using a tuple containing a ClassConverter and an Exception. I wrote it this way so that we would be able to determine the specific class that was being processed when the error occurred and the location of that class in the user's project to potentially make debugging easier. If you think this level of specificity is unnecessary then I'll change it to what you recommended

Eruanion (dismissed):

src/CTA.WebForms/FileConverters/CodeFileConverter.cs
saikiranakula-amzn: I would think "{failedTask.Item1.OriginalClassName} class " + $"located at {failedTask.Item1.FullPath}" info will be present in the stack trace. All you need to see is whether we can retrieve the stack trace from the flatten exception object, from the approach I specified above. Let me know your thoughts :)

Beau-Gosse-dev, line 85: If an exception gets thrown inside the try, will it immediately go to the catch and then stop processing the other tasks in the WhenAll? Or will it try to process all the tasks no matter what? If that case, can this catch be run many times for 1 try?

A test might help me see what is going on here if it's not too hard to write.

comsemer: `There shouldn't be any issues retrieving the stack trace from the exceptions retrieved by the `Flatten()` call, but I don't believe that `OriginalClassName` and `FullPath` would show up in the stack trace, or at least I didn't see them. To be clear, those are the original class name and path of whatever _user written_ class that is being processed, not of the converter class itself. I think I will need to add a `Flatten()` call regardless of the approach to account for potential inner aggregate exceptions, will look more into this though

comsemer: `Task.WhenAll` creates a single task that doesn't complete until it runs each of the tasks passed to it to completion regardless if one or more error out before some of the others are finished. It wraps all errors encountered into a single `AggregateException` which is passed to the `catch` block (the `catch` block only needs to be triggered once this way). One thing to note is that the original exceptions can still be accessed via the original tasks that created them, which is how we are currently accessing them in the `catch` block

src/CTA.WebForms/FileConverters/CodeFileConverter.cs
```diff
@@ -80,17 +80,27 @@ public override async Task<IEnumerable<FileInformation>> MigrateFileAsync()
...
             {
-                LogHelper.LogError(e, $"{Rules.Config.Constants.WebFormsErrorTag}Collection of migration tasks experienced 1 or more failures");
+                // NOTE: While using await does automatically unwrap AggregateExceptions when they
+                // only contain 1 exception, component tasks of Task.WhenAll() will never have their
+                // AggregateExceptions unwrapped when the WhenAll task is awaited
 
-                var failedMigrationTasks = classMigrationTasks.Where(t => t.Item2.Status != TaskStatus.RanToCompletion);
+                var failedMigrationTasks = classMigrationTasks
+                    // Select failed tasks only
+                    .Where(t => t.Item2.Status != TaskStatus.RanToCompletion)
```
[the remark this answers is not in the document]
comsemer: Good idea, I'll do this

saikiranakula-amzn (approved):

saikiranakula-amzn (approved):

Eruanion (approved):

Beau-Gosse-dev (approved):

jonlouie (approved):

Issue #511, opened 2022-01-20
Multiple semantic models returned


jonlouie:
## Describe the bug
In rare cases it is possible for multiple semantic models to be returned in the CodeConverter class. This can be prevented by searching for the semantic model via FullPath rather than RelativePath.

PR #512, opened 2022-01-20
Get semantic model with FullPath

jonlouie:
## Related issue

Closes: #511 

## Description
* Use FullPath instead of RelativePath when searching for semantic model (current search function will return multiple semantic models in rare cases where two files share the same name).

## Supplemental testing
* This fix was tested and verified manually. Unit tests might not work here because the entity to test (semantic model) is private.

---

### DIFF
--- a/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
+++ b/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
@@ -42,11 +42,11 @@ namespace CTA.WebForms.FileConverters
             try
             {
                 _fileModel = _webFormsProjectAnaylzer.AnalyzerResult.ProjectBuildResult?.SourceFileBuildResults?
-                    .Single(r => r.SourceFileFullPath.EndsWith(RelativePath))?.SemanticModel;
+                    .Single(r => r.SourceFileFullPath == FullPath)?.SemanticModel;
             }
             catch (Exception e)
             {
-                LogHelper.LogError(e, $"{Rules.Config.Constants.WebFormsErrorTag}Exception occurred when trying to retrieve semantic model for the file {RelativePath}. " +
+                LogHelper.LogError(e, $"{Rules.Config.Constants.WebFormsErrorTag}Exception occurred when trying to retrieve semantic model for the file {FullPath}. " +
                                       "Semantic Model will default to null.");
             }
 
--- a/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
+++ b/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
@@ -42,7 +42,7 @@ namespace CTA.WebForms.FileConverters
             try
             {
                 _fileModel = _webFormsProjectAnaylzer.AnalyzerResult.ProjectBuildResult?.SourceFileBuildResults?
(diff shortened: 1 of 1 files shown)


src/CTA.WebForms/FileConverters/CodeFileConverter.cs
```diff
@@ -42,11 +42,11 @@ public class CodeFileConverter : FileConverter
             try
             {
                 _fileModel = _webFormsProjectAnaylzer.AnalyzerResult.ProjectBuildResult?.SourceFileBuildResults?
-                    .Single(r => r.SourceFileFullPath.EndsWith(RelativePath))?.SemanticModel;
+                    .Single(r => r.SourceFileFullPath == FullPath)?.SemanticModel;
```
saikiranakula-amzn, line 45: use this for comparing strings to make sure we don't have any additional errors 

_fileModel = _webFormsProjectAnaylzer.AnalyzerResult.ProjectBuildResult?.SourceFileBuildResults?
                    .Single(r => r.SourceFileFullPath.Equals(fullPath, StringComparison.OrdinalIgnoreCase))?.SemanticModel;

saikiranakula-amzn (changes requested):

src/CTA.WebForms/FileConverters/CodeFileConverter.cs
jonlouie: Good catch, will make that change

src/CTA.WebForms/FileConverters/CodeFileConverter.cs
```diff
@@ -42,11 +42,11 @@ public class CodeFileConverter : FileConverter
             try
             {
                 _fileModel = _webFormsProjectAnaylzer.AnalyzerResult.ProjectBuildResult?.SourceFileBuildResults?
-                    .Single(r => r.SourceFileFullPath.EndsWith(RelativePath))?.SemanticModel;
+                    .Single(r => r.SourceFileFullPath.Equals(fullPath, StringComparison.OrdinalIgnoreCase))?.SemanticModel;
```
Beau-Gosse-dev, line 45: Just a thought: I like specifying the StringComparer, but do we actually want to ignore case here? Linux file systems are generally case sensitive. That said, I don't know if we've ever run any of this stuff on Linux anyway and if a ton of other stuff would break anyway.

Beau-Gosse-dev (approved):

comsemer (approved):

src/CTA.WebForms/FileConverters/CodeFileConverter.cs
jonlouie: For the most part we can assume that that this will be run in a Windows environment, but CTA can run on a Linux machine so I think it's an edge case we should be aware of. 

@saikiranakula-amzn What are your thoughts here?

saikiranakula-amzn: The paths we are looking here are location of source files in a project, which will follow a specific hierarchal structure, It doesn't matter whether the platform is either windows/linux

saikiranakula-amzn (approved):

Issue #513, opened 2022-01-20
When semantic model is null, classConverters collection is not initialized


jonlouie:
## Describe the bug
When semantic model is not found/null, classConverters collection is not initialized. This leads to a null reference exception when migration is attempted.

PR #514, opened 2022-01-20
Always initialize classConverters collection

jonlouie:
## Related issue

Closes: #513

## Description
* Initialize classConverters collection if semantic model is not found

## Supplemental testing
* Manually verified the expected behavior. Unable to unit test due to dependence on private members.

---

### DIFF
--- a/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
+++ b/src/CTA.WebForms/FileConverters/CodeFileConverter.cs
@@ -57,7 +57,7 @@ namespace CTA.WebForms.FileConverters
             else
             {
                 _classConverters = new List<ClassConverter>();
-                LogHelper.LogWarning("Semantic model was found so class conversion cannot take place. Returning zero class converters.");
+                LogHelper.LogWarning($"Semantic model was not found so class conversion cannot take place. Returning zero class converters for file: {FullPath}");
             }
         }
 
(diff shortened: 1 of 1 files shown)


comsemer (approved):
LGTM

Beau-Gosse-dev (approved):

src/CTA.WebForms/FileConverters/CodeFileConverter.cs
```diff
@@ -49,10 +49,15 @@ public class CodeFileConverter : FileConverter
...
+            
             if (_fileModel != null)
             {
-                _classConverters = classConverterFactory.BuildMany(RelativePath, _fileModel);
+                _classConverters = _classConverterFactory.BuildMany(RelativePath, _fileModel);
+            }
+            else
+            {
+                _classConverters = new List<ClassConverter>();
+                LogHelper.LogWarning($"Semantic model was not found so class conversion cannot take place. Returning zero class converters for file: {FullPath}");
```
saikiranakula-amzn, line 60: why this log level is warning? if there are no class convertor in the file that we support, do we log it as info?

saikiranakula-amzn (approved):

src/CTA.WebForms/FileConverters/CodeFileConverter.cs
jonlouie: I logged it as a warning because in this case, _classConverters being empty is a side effect of an error condition (missing semantic model). 

It's possible that 0 class converters would have been returned anyway (in which case nothing would be logged), but the warning signifies that particular case might be worth looking into.

saikiranakula-amzn: makes sense. Approved :)

Issue #523, opened 2022-02-02
Projects outside solution folder do not get ported


Eruanion:
## Describe the bug
Projects outside solution folder do not get ported.

### Expected behavior
All projects in the solution should attempt to port.

### Steps to Reproduce
Create a solution with a project reference outside of the solution folder (higher in tree folder hierarchy) and try to port solution.

### Desktop (please complete the following information)
 - OS: [e.g. Windows, macOS, Linux, etc]
 - Version/Distro: [e.g. 11, Big Sur, Ubuntu 18.04]

PR #524, opened 2022-02-02
Fix Projects outside solution folder do not get ported

Eruanion:
## Related issue

Closes: #523 

## Description
Projects outside solution folder do not get ported.

## Supplemental testing
All unit tests should pass including the new one.

## Additional context

---

### DIFF
--- a/tst/CTA.Rules.Test/AwsRulesBaseTest.cs
+++ b/tst/CTA.Rules.Test/AwsRulesBaseTest.cs
@@ -11,6 +11,7 @@ using System.IO.Compression;
 using System.Linq;
 using System.Reflection;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace CTA.Rules.Test
 {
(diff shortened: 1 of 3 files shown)


saikiranakula-amzn (approved):

jonlouie (dismissed):

comsemer (dismissed):

tst/CTA.Rules.Test/AwsRulesBaseTest.cs
```diff
@@ -253,7 +254,39 @@ protected string CopySolutionFolderToTemp(string solutionName, string tempDir)
...
+                    if (!Utils.IsSubPathOf(solutionDir, projPath))
+                    {
+                        string relativeSrc = Path.GetRelativePath(solutionDir, projPath);
+                        string projName = Path.GetFileName(project);
+                        string newRelDir = Path.Combine(newTempDir, relativeSrc);
+
+                        Utils.CopyFolderToTemp(projName, projPath, newRelDir);
+                    }
+                }
+            }
```
jonlouie, line 274: Correct me if I'm wrong, but my understanding here is that if what we're copying is not a solution folder (i.e. if it's a solution file), then do something different (i.e. build a folder structure that supports copying any projects that don't lie within the solution directory's direct path).

If that's the case, can we put this logic in a separate function called `CopySolutionAndProjects` or something similar? This will reduce complexity of the existing method and prevent misunderstanding of what this method does (the name assumes the input will be a directory). We may also want to add helper functions for the other stuff that's going on, for example the part where we rebuild the relative file structure for the project files.

tst/CTA.Rules.Test/CTATests.cs
```diff
@@ -210,6 +208,18 @@ public void TestMonolithReplacementsWebAPI()
...
+        public void TestProjectOutsideSolutionFolder()
+        {
+            var results = AnalyzeSolution("Application_Proj_diff_folder.sln", tempDir, downloadLocation, version);
+
+            var csProjContent1 = results.ProjectResults[0].CsProjectContent;
+            var csProjContent2 = results.ProjectResults[1].CsProjectContent;
+
+            Assert.True(csProjContent1.IndexOf(string.Concat(">", version, "<")) > 0);
+            Assert.True(csProjContent2.IndexOf(string.Concat(">", version, "<")) > 0);
+        }
```
jonlouie, line 221: nit: Can we rename the method to give more context for what the test is checking for? For example, `TestProjectsOutsideSolutionPath_Are_Ported` or similar.

Also, if the test is renamed, it may make sense to change the assertions to better describe the behavior we're looking for, for example: 
* Assert that the project files exist 
* Assert that the project file contents contain `version`

Again, these are nits but I think they will help improve readability- let me know if you think differently!

jonlouie (changes requested):

tst/CTA.Rules.Test/CTATests.cs
Eruanion: Agreed yup.

tst/CTA.Rules.Test/AwsRulesBaseTest.cs
Eruanion: I will extract the copy external projects portion to another method as you suggested.

Eruanion: Will also extract the rebuilding of the relative file structure, thanks good ideas :-)

jonlouie (approved):

saikiranakula-amzn:
Thank you Jonathan :)

PR #573, opened 2022-02-17
fix: SyntaxNodeExtension NoComments function is not removing all comments

Mingxue008:
## Action execution validation errors after ReplaceIdentifier or ReplaceMethodWithObjectAddType.

Closes: #ISSUE_NUMBER_HERE

## Description
* CTA is to validate the target code against the source code after each conversion, so to make sure target code doesn't contain any keywords that are not expected. It is done by reconstruct a syntax tree without any comments. The existing function is only removing comments from SyntaxNodes but not removing comments from SyntaxTokens. This PR is to make sure all comments or white spaces are removed properly so the validation only consider code but not comment area for any error reports.

## Supplemental testing
Source code:
Screen Shot 2022-02-16 at 10 34 58 PM

Before the change after calling NoComments function:
Screen Shot 2022-02-16 at 10 34 23 PM

After the change after calling NoComments function:
Screen Shot 2022-02-16 at 10 32 30 PM

## Additional context

---

### DIFF
--- a/src/CTA.FeatureDetection.Common/Extensions/SyntaxNodeExtensions.cs
+++ b/src/CTA.FeatureDetection.Common/Extensions/SyntaxNodeExtensions.cs
@@ -7,13 +7,21 @@ namespace CTA.FeatureDetection.Common.Extensions
 {
     public static class SyntaxNodeExtensions
     {
+        // A syntax tree has SyntaxNodes, SyntaxTokens and SyntaxTrivia, for more detail please refer to link below.
+        // https://docs.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/get-started/syntax-analysis
+        // This function is to remove all comments from the syntax tree, which it needs to iterate and replaces all
+        // SyntaxNodes and SyntaxTokens to so no comments or white spaces are in the tree anymore.
         public static SyntaxNode NoComments(this SyntaxNode node)
         {
             if (node != null)
             {
-                var childNodes = node.ChildNodes();
-                node = node.ReplaceNodes(childNodes, (node, returnedNode) => node.NoComments());
                 node = node.WithoutTrivia();
+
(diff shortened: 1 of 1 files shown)


Beau-Gosse-dev (changes requested):
Can you take the code you have in the description and turn it into a unit test?

marknfawaz (changes requested):
LGTM. Please add a test.

saikiranakula-amzn:
> Can you take the code you have in the description and turn it into a unit test? +1

jonlouie:
Agree with everyone else on the unit test(s). Otherwise looks great- thanks for the doc comment as well

Eruanion (approved):

Eruanion (approved):

src/CTA.Rules.Update/CodeReplacer.cs
```diff
@@ -251,7 +251,7 @@ public List<GenericActionExecution> ValidateActions(List<GenericActionExecution>
 
                 if (string.IsNullOrEmpty(actionValidation.CheckComments) || !bool.Parse(actionValidation.CheckComments))
                 {
-                    trimmedResult = Utils.EscapeAllWhitespace(root.NoComments().NormalizeWhitespace().ToFullString());
+                    trimmedResult = root.RemoveAllTrivia().ToFullString();
```
marknfawaz, line 254: Why are we not escaping white space anymore? As far as i know, removing trivia doesn't remove the whitespace in the code.

Mingxue008: Trivia include white spaces, empty lines and comments. Basically what we previously do here was trying to remove all trivia. That's also why the function name change regarding to this context.

marknfawaz (approved):

saikiranakula-amzn (approved):

PR #618, opened 2022-03-14
Added Template Parsing and Implemented Other Skeletons

comsemer:
Sorry this is so big, I didn't realize how much had been written until I started the pr.

## Description
* Implemented TagTemplateParser
* Implemented TagTypeConverter
* Implemented TemplateTagConverter
* Implemented TemplateCondition classes
* Added config model validation
* Added unit tests for above tasks

---


src/CTA.WebForms/Helpers/TagConversion/TagConfigParser.cs
```diff
@@ -45,9 +45,14 @@ public TagConfigParser(string configsDir)
...
-
-                    result.Add(tagName, converter);
+                    if (converter.Validate())
+                    {
+                        result.Add(tagName, converter);
+                    }
+                    else
+                    {
+                        LogHelper.LogError($"{Rules.Config.Constants.WebFormsErrorTag}Failed to validate generated converter for config at {filePath}");
+                    }
```
jonlouie, line 55: Since failed validations are considered errors, we should also log the parameters that caused the error. We can do this by either logging the errors within the converters or have `Validate()` throw an exception with a detailed error message and catch/log the exception here.

comsemer: Good catch, made the change

jonlouie (approved):
LGTM

src/CTA.WebForms/Helpers/TagConversion/TagTemplateParser.cs
```diff
@@ -1,9 +1,211 @@
...
     {
+        private CodeBehindReferenceLinkerService _codeBehindLinkerService;
+
+        public const string TargetAttributeGroup = "TargetAttribute";
+        public const string SourceAttributeGroup = "SourceAttribute";
+        public const string Context0Group = "Context0";
+        public const string Context1Group = "Context1";
+
+        /// <summary>
+        /// Regular expression that matches attribute assignments with template replacement for values.
```
Eruanion, line 23: Can we have sample of valid matches and a good explanation of what the regex string is looking for exactly?

Eruanion (approved):

PR #630, opened 2022-03-24
Webforms code behind conversion

comsemer:
Once again, very sorry for the massive PR, this is the last one though.

## Description
* Control conversion now includes code behind reference conversion and binds generated properties in the view layer
* TagTemplateParser now attempts to download config files from S3 and is more efficient all around
* Tests added for any new/modified structures
* Existing tests affected by these changes were modified

---


src/CTA.WebForms/TagConverters/TagConverter.cs
```diff
@@ -37,24 +42,86 @@ public abstract class TagConverter
...
+        /// <returns>The code behind handler type if it can be found, null otherwise</returns>
+        private protected Type GetCodeBehindHandlerType()
+        {
+            if (CodeBehindHandler == null)
+            {
+                return null;
+            }
+
+            var handlerClassName = CodeBehindHandler.Equals("Default", StringComparison.InvariantCultureIgnoreCase)
+                ? "DefaultTagCodeBehindHandler" : CodeBehindHandler;
```
jonlouie, line 101: nit: would it be safer to use nameof(DefaultTagCodeBehindHandler)?

src/CTA.WebForms/Services/CodeBehindReferenceLinkerService.cs
```diff
@@ -9,15 +17,306 @@ namespace CTA.WebForms.Services
...
+        /// code behind conversions be necessary.</param>
+        /// <param name="targetAttribute">The attribute that the conversion result will be assigned to,
+        /// if one exists, otherwise null.</param>
+        /// <param name="handler">The code behind handler to be used on this node, if one exists.</param>
+        /// <param name="token">A cancellation token for stopping processes if stall occurs.</param>
+        /// <exception cref="InvalidOperationException">Throws if view file with path <paramref name="viewFilePath"/>
+        /// has not been registered.</exception>
         /// <returns>Property binding text if a bindable property was generated, null otherwise.</returns>
-        public string GetBindingValueIfExists()
+        public async Task<string> HandleCodeBehindForAttribute(
```
jonlouie, line 48: nit: can we append "Async" to the method name?

src/CTA.WebForms/Services/CodeBehindReferenceLinkerService.cs
```diff
@@ -9,15 +17,306 @@ namespace CTA.WebForms.Services
...
+        /// Converts tag code behind references in <paramref name="classDeclaration"/> using code behind
+        /// handlers necessary for the tags found in the view file at <paramref name="viewFilePath"/>.
+        /// </summary>
+        /// <param name="viewFilePath">The view file whose tag references are to be converted.</param>
+        /// <param name="semanticModel">The semantic model that the <paramref name="classDeclaration"/> belongs to.</param>
+        /// <param name="classDeclaration">The class declaration where the code behind references will be replaced.</param>
+        /// <param name="token">A cancellation token for stopping processes if stall occurs.</param>
+        /// <returns>The modified class declaration if modifications needed to be made, otherwise
+        /// the original value of <paramref name="classDeclaration"/> is returned.</returns>
+        public async Task<ClassDeclarationSyntax> ExecuteTagCodeBehindHandlers(
```
jonlouie, line 246: nit: can we append "Async" to the method name?

src/CTA.WebForms/TagConverters/TemplateTagConverter.cs
```diff
@@ -71,7 +70,7 @@ public override void Validate()
         }
 
         /// <inheritdoc/>
-        public override void MigrateTag(HtmlNode node)
+        public override async Task MigrateTag(HtmlNode node, string viewFilePath, TagCodeBehindHandler handler, int taskId)
```
jonlouie, line 73: nit: can we append "Async" to the method name?

src/CTA.WebForms/Helpers/TagConversion/TagConfigParser.cs
```diff
@@ -18,6 +19,15 @@ public class TagConfigParser
...
         private readonly IDeserializer _deserializer;
 
+        /// <summary>
+        /// The map of known tag names to converter instances.
+        /// </summary>
+        public ConcurrentDictionary<string, TagConverter> TagConverterMap { get; }
+        /// <summary>
+        /// The list of known tag names which do not have converters.
+        /// </summary>
+        public List<string> NoConverterTags { get; }
```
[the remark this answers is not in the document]
comsemer: I think so, good catch

src/CTA.WebForms/Helpers/TagConversion/TagConfigParser.cs
```diff
@@ -49,19 +74,68 @@ public TagConfigParser(string configsDir)
...
-                    if (!result.TryAdd(converter.TagName, converter))
+                    if (!TagConverterMap.TryAdd(converter.TagName, converter))
                     {
                         LogHelper.LogError($"{Rules.Config.Constants.WebFormsErrorTag}Failed to add valid converter of type " +
                             $"{converter?.GetType().Name} for {converter?.TagName} to concurrent dictionary for config at {filePath}");
                     }
+
+                    // Even if adding to the dictionary fails, we want to return the validated converter,
+                    // the dictionary only serves to improve performance when retrieving converters
+                    return converter;
```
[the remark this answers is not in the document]
comsemer: The reason we use try add here is because its a concurrent dictionary. We need to first check if there is an existing key in the dictionary, then attempt to add it, but doing this is non-atomic. The concurrent dictionary's TryAdd basically just does the extra error handling if there was a change to the dictionary between the ContainsKey call and the accessor call that creates an invalid state. Will modify the error message to be more specific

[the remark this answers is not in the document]
jonlouie: Got it, that makes sense. Ty for the clarification

src/CTA.WebForms/ClassConverters/PageCodeBehindClassConverter.cs
```diff
@@ -50,11 +53,14 @@ public override Task<IEnumerable<FileInformation>> MigrateClassAsync()
             //     // This is so we can use ComponentBase base class
             //     .Append(Constants.BlazorComponentsNamespace);
 
+            var currentClassDeclaration = await DoTagCodeBehindConversions((ClassDeclarationSyntax)_originalDeclarationSyntax);
```
saikiranakula-amzn, line 56: always check for null before doing type casting

src/CTA.WebForms/Extensions/RegexExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public static class RegexExtensions
+    {
+        /// <summary>
+        /// Allows for regex replacements using async function delegates.
+        /// </summary>
+        /// <param name="regex">The regular expression that is being used.</param>
+        /// <param name="input">The string that <paramref name="regex"/> is being applied to.</param>
+        /// <param name="replacementFn">The function that is used for match replacement.</param>
+        /// <returns>The string <paramref name="input"/> with all matches made by <paramref name="regex"/>
```
saikiranakula-amzn, line 16: just specify return value, it is confusing when you include paramref name="input" in the comments

src/CTA.WebForms/Helpers/TagConversion/TagConfigParser.cs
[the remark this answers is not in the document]
saikiranakula-amzn: ensure the new datastructure is alos thread safe

src/CTA.WebForms/TagCodeBehindHandlers/TagCodeBehindHandler.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        /// The value of the ID attribute for the control that this handler is tasked with
+        /// handling.
+        /// </summary>
+        public string IdValue { get; }
+        /// <summary>
+        /// Conversions that have been staged for code behind references to the control
+        /// that this handler is tasked with handling.
+        /// </summary>
+        public IDictionary<SyntaxNode, SyntaxNode> StagedConversions;
```
saikiranakula-amzn, line 23: this seems little bit weird, if key and value both are syntaxNode why  not to use a hashset

src/CTA.WebForms/TagConverters/TagConverter.cs
```diff
@@ -37,24 +42,86 @@ public abstract class TagConverter
...
+        /// </summary>
+        /// <param name="idValue">The value of the ID attribute for the node which this
+        /// handler will be used on.</param>
+        /// <returns>An instance of the converter's code behind handler class if
+        /// one was specified, otherwise null.</returns>
+        /// <exception cref="InvalidOperationException">Throws if code behind handler
+        /// is specified but class couldn't be found.</exception>
+        public TagCodeBehindHandler GetCodeBehindHandlerInstance(string idValue)
+        {
+            if (CodeBehindHandler == null || idValue == null)
```
saikiranakula-amzn, line 73: use String.IsNullOrWhiteSpace(idValue) instead of just null

tst/CTA.Rules.Test/Models/WebFormsFullExpectedOutputConstants.cs
```diff
@@ -353,10 +357,10 @@ protected override void OnInitialized()
             Model = CatalogService.GetCatalogItemsPaginated(DefaultPageSize, DefaultPageIndex);
             PeopleModel = new List<People>()
             {new People(""Andy"", ""Wayne"", ""PG""), new People(""Bill"", ""Johnson"", ""SD""), new People(""Caroline"", ""Barry"", ""Manager"")};
-            PeopleGrid.DataSource = PeopleModel;
-            PeopleGrid.DataBind();
-            productList.DataSource = Model.Data;
-            productList.DataBind();
+            PeopleGrid_DataSource = PeopleModel;
+             // PeopleGrid.DataBind();
```
saikiranakula-amzn, line 361: why are these commented

comsemer: Methods are not supported for conversion, and while the DataBind() call is actually valid on BlazorWebFormsComponents components (like grid view) the method call doesn't do anything in Blazor. Basically invalid member accesses just get commented out and we convert all the rest

src/CTA.WebForms/ClassConverters/PageCodeBehindClassConverter.cs
comsemer: In this case it should be impossible for _originalDeclarationSyntax to be null because we have to use the syntax node to determine that we need to initialize the current class instance, but will add a check regardless for safety

src/CTA.WebForms/TagCodeBehindHandlers/TagCodeBehindHandler.cs
comsemer: Hash set is a better idea, will change

saikiranakula-amzn (approved):

src/CTA.WebForms/Helpers/TagConversion/TagTemplateParser.cs
```diff
@@ -233,28 +284,43 @@ private string HandleBasicReplacement(Match m, HtmlNode node)
...
+                            targetAttribute,
+                            handler,
+                            token));
 
-                // TODO: Remove below code when above code is uncommented
-                return convertedSourceValue ?? string.Empty;
+                    if (codeBehindRefBinding != null)
+                    {
+                        var x = codeBehindRefBinding;
+                    }
```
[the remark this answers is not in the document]
comsemer: Sorry this was left over from a test I believe

src/CTA.WebForms/Services/CodeBehindReferenceLinkerService.cs
```diff
@@ -9,15 +17,306 @@ namespace CTA.WebForms.Services
...
+        /// A process used to block execution the code behind class declaration of a given view file
+        /// has been registered. Assumes that the view file at <paramref name="viewFilePath"/> has already
+        /// been verified as registered.
+        /// </summary>
+        /// <param name="viewFilePath">The view file whose code behind requires registration.</param>
+        /// <param name="token">A cancellation token for stopping processes if stall occurs.</param>
+        /// <returns>A task that completes once the code behind class declaration of a given view
+        /// file has been registered.</returns>
+        private Task<bool> WaitForClassDeclarationRegistered(string viewFilePath, CancellationToken token)
+        {
```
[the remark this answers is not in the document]
comsemer: Nope, there's no awaiting going on inside so there's no need for the async modifier, we specifically want the Task to block execution in certain places

src/CTA.WebForms/Services/CodeBehindReferenceLinkerService.cs
```diff
@@ -9,15 +17,306 @@ namespace CTA.WebForms.Services
...
+        /// <param name="viewFilePath">The full path of the view file.</param>
+        /// <exception cref="InvalidOperationException">Throws if view file with path <paramref name="viewFilePath"/>
+        /// has already been registered.</exception>
+        public void RegisterViewFile(string viewFilePath)
+        {
+            if (IsViewFileRegistered(viewFilePath))
+            {
+                throw new InvalidOperationException($"{Rules.Config.Constants.WebFormsErrorTag}Failed to register view file, " +
+                    $"view file registration for path {viewFilePath} already completed");
+            }
```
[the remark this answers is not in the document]
comsemer: In theory this should never happen, but in case it does just skipping it is the most graceful way we can handle it

src/CTA.WebForms/Services/CodeBehindReferenceLinkerService.cs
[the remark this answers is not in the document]
jonlouie: Got it, thanks for the clarification

jonlouie (approved):

PR #638, opened 2022-03-31
Add source versions to CTA framework and metrics

Eruanion:
## Related issue

Closes: #639 

## Description
Currently CTA is passed source version from PA but loses track of it within its models.

## Supplemental testing
Pass all unit tests

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Metrics/MetricsTransformer.cs
+++ b/src/CTA.Rules.Metrics/MetricsTransformer.cs
@@ -54,9 +54,15 @@ namespace CTA.Rules.Metrics
             var targetVersions = projectResult.TargetVersions;
             var sourceVersions = projectResult.SourceVersions;
             var targetVersionMetrics = new List<TargetVersionMetric>();
+
             for (int i = 0; i < targetVersions.Count; i++)
             {
-                targetVersionMetrics.Add(new TargetVersionMetric(context, targetVersions[i], projectFile, sourceVersions[i]??null));
+                string sourceVersion = null;
+                if (targetVersions.Count == sourceVersions.Count)
+                {
+                    sourceVersion = sourceVersions[i];
+                }
+                targetVersionMetrics.Add(new TargetVersionMetric(context, targetVersions[i], projectFile, sourceVersion));
             }
 
             return targetVersionMetrics;
--- a/src/CTA.Rules.Metrics/MetricsTransformer.cs
+++ b/src/CTA.Rules.Metrics/MetricsTransformer.cs
@@ -52,10 +52,11 @@ namespace CTA.Rules.Metrics
         {
             var projectFile = projectResult.ProjectFile;
             var targetVersions = projectResult.TargetVersions;
+            var sourceVersions = projectResult.SourceVersions;
             var targetVersionMetrics = new List<TargetVersionMetric>();
-            foreach (var targetVersion in targetVersions)
+            for (int i = 0; i < targetVersions.Count; i++)
             {
-                targetVersionMetrics.Add(new TargetVersionMetric(context, targetVersion, projectFile));
+                targetVersionMetrics.Add(new TargetVersionMetric(context, targetVersions[i], projectFile, sourceVersions[i]??null));
             }
 
             return targetVersionMetrics;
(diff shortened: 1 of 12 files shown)


src/CTA.Rules.Metrics/MetricsTransformer.cs
```diff
@@ -52,10 +52,11 @@ internal static IEnumerable<TargetVersionMetric> TransformTargetVersions(Metrics
         {
             var projectFile = projectResult.ProjectFile;
             var targetVersions = projectResult.TargetVersions;
+            var sourceVersions = projectResult.SourceVersions;
             var targetVersionMetrics = new List<TargetVersionMetric>();
-            foreach (var targetVersion in targetVersions)
+            for (int i = 0; i < targetVersions.Count; i++)
             {
-                targetVersionMetrics.Add(new TargetVersionMetric(context, targetVersion, projectFile));
+                targetVersionMetrics.Add(new TargetVersionMetric(context, targetVersions[i], projectFile, sourceVersions[i]??null));
```
jonlouie, line 59: It doesn't look like we enforce `projectResult.TargetVersions` and `projectResult.SourceVersions` to have the same number of elements.

Can we have a quick check or process to reconcile when the two data structures are different lengths?

src/CTA.Rules.Models/ProjectResult.cs
```diff
@@ -9,6 +9,7 @@ public class ProjectResult
         public List<PackageAction> UpgradePackages { get; set; }
         public List<PackageAction> ActionPackages { get; set; }
         public List<string> TargetVersions { get; set; }
+        public List<string> SourceVersions { get; set; }
```
saikiranakula-amzn, line 12: source version can just be string, since each project have only one framework version, but no harm in keep it

src/CTA.Rules.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30309.148
+# Visual Studio Version 17
```
saikiranakula-amzn, line 3: can I get some context here, I don't know why we updated this version. Curious to know :)

src/CTA.Rules.Models/ProjectResult.cs
Eruanion: Yeah I wanted to stay consistent, definitely agree with the fact they can both just be strings but I think Mark had ideas on how to use this in the future which is why he went with a list instead. I'll keep it like this to match targetVersions for now.

src/CTA.Rules.sln
Eruanion: Oh this was accidental :-) However not a bad idea to upgrade to latest :-)
Let me know if you have any reasons not to check this in.

src/CTA.Rules.Metrics/MetricsTransformer.cs
Eruanion: Added a check.

skakula91 (approved):

src/CTA.Rules.ProjectFile/ProjectFileCreator.cs
```diff
@@ -99,6 +100,27 @@ public class ProjectFileCreator
             }
         }
 
+        public ProjectFileCreator(string projectFile, List<string> targetVersions, Dictionary<string, string> packages,
```
comsemer, line 103: Might be cleaner to chain constructors here since everything past `_sourceVersions = sourceVersions;` looks to be the same as the content for the constructor above this one, but not a big deal either way.

comsemer (commented):
Just one minor comment, otherwise looks good

src/CTA.Rules.ProjectFile/ProjectFileCreator.cs
Eruanion: ouh good catch, fixed it. Thanks :-)

comsemer (approved):

saikiranakula-amzn (approved):

jonlouie (approved):

saikiranakula-amzn (approved):

jonlouie (approved):

PR #646, opened 2022-04-22
Support remove using directive actions inside namespace block

Mingxue008:
## Related issue
Some developers choose to write using directives inside the namespace block, and we are not applying our remove actions for those situations.

Closes: #ISSUE_NUMBER_HERE

## Description
* We need to handle removing using directive actions in both CompilationUnit and inside Namespace. 
* Regarding add using directives, we are only adding them inside CompilationUnit as our convention. 

## Supplemental testing
Added unit tests to cover the changes in this PR.
Adjusted the build error count based on below changes by updating our BuidableWebApi project at https://github.com/marknfawaz/TestProjects for the test case:
`namespace BuildableWebApi.Controllers
{
    using System.Web.Http;
    using System.Web.Http.OData;
}`
And this project is covered in our test cases.

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Actions/NamespaceActions.cs
+++ b/src/CTA.Rules.Actions/NamespaceActions.cs
@@ -19,21 +19,12 @@ namespace CTA.Rules.Actions
             return RenameNamespace;
         }
 
-        public Func<SyntaxGenerator, NamespaceDeclarationSyntax, NamespaceDeclarationSyntax> GetAddDirectiveAction(string @namespace)
-        {
-            NamespaceDeclarationSyntax AddDirective(SyntaxGenerator syntaxGenerator, NamespaceDeclarationSyntax node)
-            {
-                var allUsings = node.Usings;
-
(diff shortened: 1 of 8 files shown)


jonlouie (commented):
Can we add a tests to verify these actions work as intended? It would also highlight the cons referenced on `src/CTA.Rules.RuleFiles/RulesFileParser.cs, line 463`

Mingxue008:
> Can we add a tests to verify these actions work as intended? It would also highlight the cons referenced on `src/CTA.Rules.RuleFiles/RulesFileParser.cs, line 463`

@jonlouie, thanks for catching those comments. I have updated it so they are not confusing. Basically I am only supporting Remove using directives inside the namespace block, so there is no Cons with this approach any more.
Regarding the test cases, it's added with the PR here: https://github.com/marknfawaz/TestProjects/pull/71

Mingxue008:
@jonlouie The test cases were added, please let me know if anything else is needed.

jonlouie (approved):

PR #647, opened 2022-04-22
Add CastExpressionSyntax into Identifier types

Mingxue008:
## Related issue
When porting code, we want to replace an identifier but failed to do so because it's of CastExpressionSyntax.

Closes: #ISSUE_NUMBER_HERE

## Description
* Added CastExpressionSyntax into identifier types so it is included in the replace actions if needed.

## Supplemental testing
The test case is added into BuildableWebApi project at https://github.com/marknfawaz/TestProjects
'return (IHttpActionResult)Ok();'
And this project is covered in our test cases.

## Additional context

---

### DIFF
--- a/src/CTA.Rules.Update/ActionsRewriter.cs
+++ b/src/CTA.Rules.Update/ActionsRewriter.cs
@@ -31,7 +31,9 @@ namespace CTA.Rules.Update.Rewriters
             typeof(TypeParameterListSyntax),
             typeof(ParameterSyntax),
             typeof(ObjectCreationExpressionSyntax),
-            typeof(QualifiedNameSyntax)};
+            typeof(QualifiedNameSyntax),
+            typeof(CastExpressionSyntax),
+        };
 
         public ActionsRewriter(SemanticModel semanticModel, SemanticModel preportSemanticModel, SyntaxGenerator syntaxGenerator, string filePath, List<GenericAction> allActions)
         {


src/CTA.Rules.Update/ActionsRewriter.cs
```diff
@@ -31,7 +31,9 @@ public class ActionsRewriter : CSharpSyntaxRewriter
             typeof(TypeParameterListSyntax),
             typeof(ParameterSyntax),
             typeof(ObjectCreationExpressionSyntax),
-            typeof(QualifiedNameSyntax)};
+            typeof(QualifiedNameSyntax),
+            typeof(CastExpressionSyntax),
```
Eruanion, line 35: There is a similar list in Codelyzer that also needs to be updated which does a similar work but allows codelyzer to recognize these nodes.

Mingxue008: Thanks for the note, let me do another PR for Codelyzer then.

jonlouie: @Eruanion I believe the part of the code that uses this identifier list is looking Roslyn's syntax tree as opposed to Codelyzer's syntax tree. We might not require changes in Codelyzer.

@Mingxue008 Can you add a unit test to verify that adding `CastExpressionSyntax` to the identifier list works as intended?

Mingxue008: PR is created in Codelyzer https://github.com/aws/codelyzer/pull/170

Mingxue008: @jonlouie, that's good points. @Eruanion, please help me understand that Codelyzer change a bit as well.
Regarding the test case, it's added in another repo:
https://github.com/marknfawaz/TestProjects/pull/71
Do you mean to add a unit test in CTA instead?

jonlouie: @Mingxue008 Yes, I meant to suggest adding a test in CTA. Feel free to merge in https://github.com/marknfawaz/TestProjects/pull/71, then the code changes will be accessible in the CTA tests. Final step is to update or add a CTA unit test to verify we get the expected changes.

Mingxue008: @jonlouie, I thought our existing test [here ](https://github.com/aws/cta/blob/2e27c744fca09a97861017cd645f86531ae6c28e/tst/CTA.Rules.Test/PortCoreTests.cs#:~:text=public%20void%20TestBuildableWebApiSolution(string%20version)) is good enough to cover the tests? Let's follow up offline for the unit tests for more detail, it's not very clear to me what unit test I should cover for this change specifically.

Eruanion: So without the codelyzer change, we will not correctly label the nodes in the json output of the codelyzer so we would never end up with a CastExpressionSyntax node in CTA.

Eruanion (approved):

jonlouie (approved):

src/CTA.Rules.Update/ActionsRewriter.cs
jonlouie: Per our offline discussion, these cases were sufficiently tested. Codelyzer will still be updated for maintenance and consistency reasons.

PR #664, opened 2022-05-06
Added null check for class baselist dereferencing

comsemer:
## Description
* Seems like class declaration base list is sometimes null, added check when dereferencing for loop
* Should fix https://sim.amazon.com/issues/V537759661

---

### DIFF
--- a/src/CTA.Rules.Analysis/RulesAnalysis.cs
+++ b/src/CTA.Rules.Analysis/RulesAnalysis.cs
@@ -201,7 +201,7 @@ namespace CTA.Rules.Analyzer
                                 }
 
                                 token = null;
-                                foreach (string interfaceName in classType.BaseList)
+                                foreach (string interfaceName in classType.BaseList ?? Enumerable.Empty<string>())
                                 {
                                     var baseListToken = new ClassDeclarationToken() { FullKey = interfaceName };
                                     _rootNodes.Classdeclarationtokens.TryGetValue(baseListToken, out token);


src/CTA.Rules.Analysis/RulesAnalysis.cs
```diff
@@ -201,7 +201,7 @@ private bool AnalyzeChildren(FileActions fileAction, UstList<UstNode> children,
                                 }
 
                                 token = null;
-                                foreach (string interfaceName in classType.BaseList)
+                                foreach (string interfaceName in classType.BaseList ?? Enumerable.Empty<string>())
```
saikiranakula-amzn, line 204: why Enumerable.Empty<string> ? we can skip this block if the classType.BaseList is empty?

Mingxue008: piggyback on this, could you add some comment explaining the change inline? In what situation this could be null for example

comsemer: Using Enumerable.Empty will result in the block being skipped. I just felt it was more concise like this than wrapping the whole block in an if statement.

ClassDeclaration is a Codelyzer class, after looking at the code for a while it seems like the only explanation is that sometimes when a class has an empty base list, rather than having an empty collection here, the value is just null. Will add a comment explaining this.

Mingxue008:
Just noticed another thing from this PR, that you posted the SIM ticket link in the description area. Just curious if that is the recommended way? I had concerns before to expose our SIM link in the public Github PRs like this, but I could see the benefit of linking those together for better bookkeeping.

comsemer:
> Just noticed another thing from this PR, that you posted the SIM ticket link in the description area. Just curious if that is the recommended way? I had concerns before to expose our SIM link in the public Github PRs like this, but I could see the benefit of linking those together for better bookkeeping.

src/CTA.Rules.Analysis/RulesAnalysis.cs
Mingxue008: Thanks for the update. Could you elaborate a bit on the logic here as well since this change will only skip the foreach() section, but would continue with the next AnalyzeChildren attempt. Is that expected or we should break right after we know the classType.BaseList is null?

saikiranakula-amzn (approved):

Mingxue008 (approved):
Thanks for the explanations, looks good to me.

PR #700, opened 2022-05-19
Applied fix for cascading task cancellations

comsemer:
## Description
* Updated task manager service to use delays before task cancellations as was originally intended
* Fixed some bugs in task manager that could potentially lead to false positive stall state

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.WebForms/Services/TaskManagerService.cs
+++ b/src/CTA.WebForms/Services/TaskManagerService.cs
@@ -19,14 +19,16 @@ namespace CTA.WebForms.Services
         public enum TaskState { Active, Waiting }
         public enum TaskStallTimeout { None = 0, Short = 100, Medium = 1000, Long = 10000 }
 
-        private IDictionary<int, ManagedTask> _managedTasks;
+        private readonly IDictionary<int, ManagedTask> _managedTasks;
+        private readonly TaskStallTimeout _stallTimeout;
         private bool _stalled;
         private int _stallOccurrenceNumber;
         private int _nextAvailableTaskId;
 
-        public TaskManagerService()
+        public TaskManagerService(TaskStallTimeout stallTimeout = TaskStallTimeout.Short)
         {
             _managedTasks = new Dictionary<int, ManagedTask>();
+            _stallTimeout = stallTimeout;
         }
 
         public int RegisterNewTask()
@@ -52,10 +54,10 @@ namespace CTA.WebForms.Services
             try
             {
                 result = await func(token);
-                managedTask.SetActive();
             }
             finally
             {
+                managedTask.SetActive();
                 UpdateStallState();
             }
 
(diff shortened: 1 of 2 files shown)


src/CTA.WebForms/Services/TaskManagerService.cs
```diff
@@ -19,14 +19,16 @@ public class TaskManagerService
         public enum TaskState { Active, Waiting }
         public enum TaskStallTimeout { None = 0, Short = 100, Medium = 1000, Long = 10000 }
```
jonlouie, line 20: Can we rename this to include the unit of time? E.g. `TaskStallTimeoutInSeconds`/`Ms`

A comment with unit of time would also be helpful.

src/CTA.WebForms/Services/TaskManagerService.cs
```diff
@@ -93,27 +95,24 @@ private void UpdateStallState()
...
-            // if (_stalled && _stallOccurrenceNumber == currentStallOccurrence)
-            // {
-            var oldestTask = _managedTasks.Values.OrderBy(managedTask => managedTask.LastStatusChange).FirstOrDefault();
-
-            if (oldestTask != null)
+            if (_stalled && _stallOccurrenceNumber == currentStallOccurrence)
             {
-                oldestTask.CancelTask();
+                var oldestTask = _managedTasks.Values.OrderBy(managedTask => managedTask.LastStatusChange).FirstOrDefault();
+
```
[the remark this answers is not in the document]
comsemer: In general, calls to `TryResolveStall` should be rare, and not every call to `TryResolveStall` will even reach the `OrderBy` because we expect the stall to be resolved naturally during the stall timeout period. My concern with using a `PriorityQueue` would be that our lookups by task id would no longer be O(1) in that data structure, and those calls will be a lot more frequent. We could maintain a parallel `PriorityQueue` data structure for stall resolution, but I wouldn't want to change `_managedTasks` from being a `Dictionary`.

src/CTA.WebForms/Services/TaskManagerService.cs
comsemer: This is a good idea, will add

src/CTA.WebForms/Services/TaskManagerService.cs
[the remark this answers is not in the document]
jonlouie: Thanks for the info. I agree, keeping it as a `Dictionary` seems like the right choice.

jonlouie (dismissed):

src/CTA.WebForms/Services/TaskManagerService.cs
```diff
@@ -19,14 +19,16 @@ public class TaskManagerService
         public enum TaskState { Active, Waiting }
         public enum TaskStallTimeout { None = 0, Short = 100, Medium = 1000, Long = 10000 }
 
-        private IDictionary<int, ManagedTask> _managedTasks;
+        private readonly IDictionary<int, ManagedTask> _managedTasks;
```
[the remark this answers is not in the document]
comsemer: Good catch

jonlouie (dismissed):

Eruanion (approved):

src/CTA.WebForms/Services/TaskManagerService.cs
```diff
@@ -17,24 +18,34 @@ public class TaskManagerService
         private const string UnstalledLogAction = "Exited Stall";
 
         public enum TaskState { Active, Waiting }
-        public enum TaskStallTimeout { None = 0, Short = 100, Medium = 1000, Long = 10000 }
+        // Predefined timeout periods (in milliseconds) for when the application enters a stall state
+        public enum TaskStallTimeoutMs { None = 0, Short = 100, Medium = 1000, Long = 10000 }
```
saikiranakula-amzn, line 22: how did you come up with these numbers?

src/CTA.WebForms/Services/TaskManagerService.cs
```diff
@@ -93,27 +106,24 @@ private void UpdateStallState()
...
         }
 
-        private void TryResolveStall()
+        private async void TryResolveStall()
         {
-            // NOTE: Commented out for use later, currently
-            // configured to immediately cancel oldest task
-            // on stall
+            var currentStallOccurrence = _stallOccurrenceNumber;
+            await Task.Delay((int)_stallTimeoutMs);
```
saikiranakula-amzn, line 112: this will hit application performance? why we need to delay the task?

comsemer: Also, if the stall resolves during the wait time, it should actually have no impact on performance at all since the application continues to run normally during this delay

src/CTA.WebForms/Services/TaskManagerService.cs
comsemer: Just used my judgement on what I thought would be a minimum "safe" timeout before cancelling tasks and used that as the value for small, then increased by a factor of 10 for medium and large

PR #759, opened 2022-06-14
fix: feature detection visual basic

saikiranakula-amzn:
## Related issue

Closes: #ISSUE_NUMBER_HERE

## Description
* Feature detection identifies the project type based on the references and project files

## Supplemental testing

## Additional context

---

### DIFF
--- a/src/CTA.FeatureDetection.Common/Extensions/ProjectWorkspaceQueries.cs
+++ b/src/CTA.FeatureDetection.Common/Extensions/ProjectWorkspaceQueries.cs
@@ -41,6 +41,18 @@ namespace CTA.FeatureDetection.Common.Extensions
                 .Any(c => c.BaseTypeOriginalDefinition == typeOriginalDefinition);
 
         /// <summary>
+        /// Determines if a ProjectWorkspace declares a class with a specified base type for vb sln
+        /// </summary>
+        /// <param name="project">ProjectWorkspace to search</param>
+        /// <param name="typeOriginalDefinition">Original Definition of the base type being searched for</param>
+        /// <returns>Whether or not a class with the specified base type is declared in the project</returns>
+        public static bool DeclaresClassBlocksWithBaseType(this ProjectWorkspace project, string typeOriginalDefinition)
+            => project.SourceFileResults
+                .SelectMany(n => n.AllClassBlocks())
+                .Any(c => c.BaseTypeOriginalDefinition == typeOriginalDefinition);
+
+
+        /// <summary>
         /// Gets all class declaration nodes in a ProjectWorkspace
         /// </summary>
         /// <param name="project">ProjectWorkspace to search</param>
--- a/src/CTA.FeatureDetection.ProjectType/CompiledFeatures/VBClassLibraryFeature.cs
+++ b/src/CTA.FeatureDetection.ProjectType/CompiledFeatures/VBClassLibraryFeature.cs
@@ -1,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+using Codelyzer.Analysis;
+using CTA.FeatureDetection.Common.Extensions;
+using CTA.FeatureDetection.Common.Models.Features.Base;
+using CTA.Rules.Config;
+using Microsoft.Build.Construction;
+
+namespace CTA.FeatureDetection.ProjectType.CompiledFeatures
+{
+    public class VBClassLibraryFeature : CompiledFeature
+    {
+        /// <summary>
+        /// Determines if a project is an VB class library project by looking at file extensions and nuget references
+        /// </summary>
+        /// <param name="analyzerResult"></param>
+        /// <returns>Whether a project is an VB class library project or not</returns>
+        public override bool IsPresent(AnalyzerResult analyzerResult)
+        {
+            var project = analyzerResult.ProjectResult;
+            var isVBProj = project.ContainsFileWithExtension(Constants.VbClassExtension, true) && project.ProjectFilePath.EndsWith(Constants.VbProjExtension)
+                            &&!(project.ContainsDependency(Constants.MvcReferenceIdentifier) && project.DeclaresClassBlocksWithBaseType(Constants.MvcControllerOriginalDefinition))
+                            &&!((project.ContainsNugetDependency(Constants.WebApiNugetReferenceIdentifier) || project.ContainsDependency(Constants.WebApiReferenceIdentifier))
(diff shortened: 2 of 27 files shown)


src/CTA.FeatureDetection.ProjectType/CompiledFeatures/VBClassLibraryFeature.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        /// <summary>
+        /// Determines if a project is an VB class library project by looking at file extensions and nuget references
+        /// </summary>
+        /// <param name="analyzerResult"></param>
+        /// <returns>Whether a project is an VB class library project or not</returns>
+        public override bool IsPresent(AnalyzerResult analyzerResult)
+        {
+            var project = analyzerResult.ProjectResult;
+            var isVBProj = project.ContainsFileWithExtension(Constants.VbClassExtension, true) && project.ProjectFilePath.EndsWith(Constants.VbProjExtension)
```
cslong, line 25: should we split out "isVBProj" and the feature specifically? especially if the .vbproj stuff is reused in the others?

src/CTA.FeatureDetection.ProjectType/CompiledFeatures/VBClassLibraryFeature.cs
```diff
@@ -0,0 +1,48 @@
...
+                            &&!(project.ContainsDependency(Constants.MvcReferenceIdentifier) && project.DeclaresClassBlocksWithBaseType(Constants.MvcControllerOriginalDefinition))
+                            &&!((project.ContainsNugetDependency(Constants.WebApiNugetReferenceIdentifier) || project.ContainsDependency(Constants.WebApiReferenceIdentifier))
+                            && project.DeclaresClassBlocksWithBaseType(Constants.WebApiControllerOriginalDefinition));
```
cslong, line 28: Just double checking we want true for WebApiControllerOriginalDefinition in the ClassLibraryFeature class.

cslong (commented):
is it possible that we can refactor the boolean logic for the feature detection so it's shared between the VB stuff and the C#? I'm not sure since I'm not familiar, but it seems like it would be the same.

src/CTA.FeatureDetection.ProjectType/CompiledFeatures/VBClassLibraryFeature.cs
saikiranakula-amzn: ! condition for the api check, since api, webforms and mvc are also library projects as output type

src/CTA.FeatureDetection.ProjectType/CompiledFeatures/VBClassLibraryFeature.cs
saikiranakula-amzn: sure I can make that change in the subsequent PR

src/CTA.FeatureDetection.ProjectType/CompiledFeatures/VBClassLibraryFeature.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml.Linq;
```
Beau-Gosse-dev, line 6: Try to not include using statements for things we're not using.

src/CTA.Rules.PortCore/SolutionPort.cs
```diff
@@ -477,6 +477,22 @@ internal ProjectType GetProjectType(FeatureDetectionResult projectTypeFeatureRes
             {
                 return ProjectType.WCFClient;
             }
+            else if(projectTypeFeatureResult.IsVBNetMvcProject())
+            {
+                return ProjectType.VBNetMvc;
+            }
+            else if (projectTypeFeatureResult.IsVBWebFormsProject())
```
Beau-Gosse-dev, line 484: I could be misreading the logic, but won't this never get hit since the condition for non-VB projects doesn't check for C# or some other language. So the AspNetWebFormsFeature will return true before it gets to check if it's also VB specific

src/CTA.FeatureDetection.ProjectType/Constants.cs
```diff
@@ -31,6 +31,10 @@ public class Constants
         public const string WCFServiceCodeFeatureName = "CoreWCFServiceCodeFeature";
         public const string WCFClientFeatureName = "WCFClientFeature";
         public const string WCFServiceHostFeatureName = "WCFServiceHostFeature";
+        public const string VBClassLibraryFeatureName = "VBClassLibraryFeature";
+        public const string VBWebApiFeatureName = "VBWebApiFeature";
+        public const string VBWebFormsFeatureName = "VBWebFormsFeature";
```
Beau-Gosse-dev, line 36: Should we be using nameof(VBWebFormsFeature) for these? Maybe don't even put them in the constants file and just use nameof(VBWebFormsFeature) where they are needed.

src/CTA.Rules.PortCore/SolutionPort.cs
Beau-Gosse-dev: If this is actually an issue, it would be worth checking all the other VB types. Also, is there a test that covers this case?

Beau-Gosse-dev (changes requested):
I haven't gone through everything. Will do the rest tomorrow morning.

src/CTA.FeatureDetection.Common/Extensions/ProjectWorkspaceQueries.cs
```diff
@@ -40,6 +40,18 @@ public static bool DeclaresClassWithBaseType(this ProjectWorkspace project, stri
                 .SelectMany(n => n.AllClasses())
                 .Any(c => c.BaseTypeOriginalDefinition == typeOriginalDefinition);
 
+        /// <summary>
+        /// Determines if a ProjectWorkspace declares a class with a specified base type for vb sln
```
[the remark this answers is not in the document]
saikiranakula-amzn: yes this is specific to vb

src/CTA.FeatureDetection.ProjectType/Constants.cs
saikiranakula-amzn: good point, but making them consistent with existing enums

src/CTA.Rules.PortCore/SolutionPort.cs
saikiranakula-amzn: that is a good point actually, will change the ordering

Beau-Gosse-dev (approved):

PR #1172, opened 2023-09-19
Handle null values regarding GetAttributeValue

jonajian:
## Related issue

Closes: none(?)

## Description
* Adds null checks to usages of GetAttributeValue

---

### DIFF
--- a/src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
+++ b/src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
@@ -137,7 +137,8 @@ namespace CTA.Rules.Actions
         private void HandleHttpRedirect(XElement httpRedirectElement)
         {
             StringBuilder sb = new StringBuilder("new RewriteOptions()");
-            bool isRedirectionEnabled = bool.Parse(GetAttributeValue(httpRedirectElement, Constants.Enabled));
+            string? redirectionString = GetAttributeValue(httpRedirectElement, Constants.Enabled);
+            bool isRedirectionEnabled = redirectionString is not null && bool.Parse(redirectionString);
             if (!isRedirectionEnabled)
                 return;
 
@@ -184,7 +185,7 @@ namespace CTA.Rules.Actions
                 {
                     var handlerName = GetAttributeValue(handler, Constants.Type);
                     var handlerPath = GetAttributeValue(handler, Constants.PathAttribute, "*");
-                    if(!ServerConfigTemplates.UnsupportedHandlers.Contains(handlerName.Trim()))
+                    if(handlerName is not null && handlerPath is not null && !ServerConfigTemplates.UnsupportedHandlers.Contains(handlerName.Trim()))
                         _middleWareExpressions[ServerConfigTemplates.ConfigureMiddlewareMethod].Add(string.Format(ServerConfigTemplates.ConfigurationExpressionTemplates[WebServerConfigAttributes.Handlers.ToString()], handlerPath, handlerName));
                 }
             }
@@ -207,8 +208,9 @@ namespace CTA.Rules.Actions
                         if (mimeType.Name.ToString().Trim().Equals(Constants.Add))
                         {
                             var mimeTypeSupported = GetAttributeValue(mimeType, Constants.MimeType);
-                            bool isMimeTypeEnabled = bool.Parse(GetAttributeValue(mimeType, Constants.Enabled));
-                            if (isMimeTypeEnabled)
+                            string? mimeTypeEnabledString = GetAttributeValue(mimeType, Constants.Enabled);
+                            bool isMimeTypeEnabled = mimeTypeEnabledString is not null && bool.Parse(mimeTypeEnabledString);
+                            if (mimeTypeSupported is not null && isMimeTypeEnabled)
                             {
                                 mimeTypeList.Add(mimeTypeSupported);
                             }
@@ -241,7 +243,8 @@ namespace CTA.Rules.Actions
                 // window authentication
                 if (authenticationType.Name.ToString().Equals(WebServerConfigAttributes.WindowsAuthentication.ToString(), System.StringComparison.OrdinalIgnoreCase))
                 {
-                    bool isAuthEnabled = bool.Parse(GetAttributeValue(authenticationType, Constants.Enabled, ""));
+                    string? authEnabledString = GetAttributeValue(authenticationType, Constants.Enabled, "");
+                    bool isAuthEnabled = authEnabledString is not null && bool.Parse(authEnabledString);
                     if (isAuthEnabled)
                     {
                         string windowsAuthentication = WebServerConfigAttributes.WindowsAuthentication.ToString();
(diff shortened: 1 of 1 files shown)


src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
```diff
@@ -137,7 +137,8 @@ private void HandleRequestFiltering(XElement requestFilteringElement)
         private void HandleHttpRedirect(XElement httpRedirectElement)
         {
             StringBuilder sb = new StringBuilder("new RewriteOptions()");
-            bool isRedirectionEnabled = bool.Parse(GetAttributeValue(httpRedirectElement, Constants.Enabled));
```
jonlouie, line 140: Do we know if `bool.Parse()` throws if the argument is null?

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
```diff
@@ -184,7 +185,7 @@ private void HandleHandlers(XElement httpHandlerElement)
                 {
                     var handlerName = GetAttributeValue(handler, Constants.Type);
                     var handlerPath = GetAttributeValue(handler, Constants.PathAttribute, "*");
-                    if(!ServerConfigTemplates.UnsupportedHandlers.Contains(handlerName.Trim()))
+                    if(handlerName is not null && !ServerConfigTemplates.UnsupportedHandlers.Contains(handlerName.Trim()))
```
jonlouie, line 188: `handlerPath` can also be null

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
```diff
@@ -207,8 +208,9 @@ private void HandleHttpCompression(XElement httpCompressionElement)
                         if (mimeType.Name.ToString().Trim().Equals(Constants.Add))
                         {
                             var mimeTypeSupported = GetAttributeValue(mimeType, Constants.MimeType);
-                            bool isMimeTypeEnabled = bool.Parse(GetAttributeValue(mimeType, Constants.Enabled));
```
jonlouie, line 210: Same as above, please confirm the behavior if `null` is passed into `bool.Parse()`

jonlouie (changes requested):
Left a few questions/comments. Please take another scan through and ensure we're handling returned values from all `GetAttributeValue ()`. In general, if we get any nulls, we want to convert them to empty strings or skip the action completely.

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
jonajian: `mimeType` will never be null here, else it would have error'd out in line 207

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
jonajian: I will fix this. I thought `string.Format` could handle null inputs, but it cannot.

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
jonajian: yes, bool.Parse() has a check in it that throws an exception if the argument is null. I am assuming that I do not need to check if the `XElement` input is null, because that pattern (not checking if the input arg is null) is repeated across all of the `Handle` functions.

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
jonlouie: The result of `GetAttributeValue()` is passed in; is that value guaranteed to never be null?

jonajian: Hmm...so `GetAttributeValue` is called twice, both times with `mimeType` as the input. `mimeType` cannot be null at this point in the code, so there is no check.

As for the output of `GetAttributeValue`:
The first time is for `var mimeTypeSupported`, and the null check for that is on line 213.
The second time is for `string? mimeTypeEnabledString`, and the null check for that is on line 212.

jonlouie: Got it. Not sure if I misread the first time or if there was a change on the follow-up commit, but it looks ok to me now.

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
jonajian: Unless it is possible for `httpRedirectElement` to not be null, but for `GetAttributeValue` to still return null...

jonajian: Wait but I'm already handling it for that case...

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
jonajian: I think for some of these, you might have been looking at the old version, the "pre-change" version

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
jonlouie: Correct. We ensure the XElement object is not null before calling any of these `Handle*` methods. 

For future reference, it is generally easier to test/maintain code if parameter validation is handled by the receiving method rather than by the caller. But there is no hard and fast rule saying one way or the other needs to be done.

jonlouie: > Wait but I'm already handling it for that case... 
Very possible, my mistake.

jonlouie (approved):

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
```diff
@@ -137,7 +137,8 @@ private void HandleRequestFiltering(XElement requestFilteringElement)
         private void HandleHttpRedirect(XElement httpRedirectElement)
         {
             StringBuilder sb = new StringBuilder("new RewriteOptions()");
-            bool isRedirectionEnabled = bool.Parse(GetAttributeValue(httpRedirectElement, Constants.Enabled));
+            string? redirectionString = GetAttributeValue(httpRedirectElement, Constants.Enabled);
```
saikiranakula-amzn, line 140: string by default can have a null value

src/CTA.Rules.Actions/ActionHelpers/ServerConfigMigrate.cs
```diff
@@ -366,7 +369,7 @@ private void AddComments()
             _startupRoot = CommentHelper.AddCSharpComment(_startupRoot, sb.ToString());
         }
 
-        private string GetAttributeValue(XElement element, string attributeName, string splitChar = null)
+        private string? GetAttributeValue(XElement element, string attributeName, string splitChar = null)
```
saikiranakula-amzn, line 372: same here. you don't need nullable on the string

